PR #23, opened 2019-11-03
Introduces the Bollinger Bands indicator

AndreMiras:
Disclaimer: I suck at trading and TA, but I can code (sometimes) :smile: 

These simple indicators are sometimes used for TA:
https://en.wikipedia.org/wiki/Bollinger_Bands

I also added some rather simple unit tests for the indicator. It might feel overkill, but better safe than sorry. I struggled a bit to know where to put the code. I feel it should be reorganized a bit. Like the indicators should have a dedicated module. Same thing for strategies and datasets. Then I we could also have a unit test file per module. That might be out of the scope of this pull request, but I'm happy to discuss it later and help with that.
I've implemented a simple strategy on top of it and I can pull request it later if there's some interest.
See it in action here: https://nbviewer.jupyter.org/urls/gitlab.com/andre.miras/quant_experiments/raw/develop/bollinger_bands.ipynb

### DIFF
--- a/backtesting/test/__init__.py
+++ b/backtesting/test/__init__.py
@@ -21,3 +21,29 @@ def SMA(arr: pd.Series, n: int) -> pd.Series:
     Returns `n`-period simple moving average of array `arr`.
     """
     return pd.Series(arr).rolling(n).mean()
+
+
+def standard_deviation(arr: pd.Series, n: int) -> pd.Series:
+    """Returns `n`-period standard deviation of array `arr`."""
+    return pd.Series(arr).rolling(n).std()
+
+
+def bollinger_bands(arr: pd.Series, n: int, std_mul: float) -> pd.Series:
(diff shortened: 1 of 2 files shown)


kernc:
This package is specifically _not_ meant to replace an indicator library such as [TA-LIB](https://github.com/mrjbq7/ta-lib), [Tulipy](https://tulipindicators.org/), [ta](https://github.com/bukosabino/ta) ...

`SMA()` is there for tests and examples. Committing to anything more sophisticated, I think, is an overkill.

AndreMiras:
I missed that, clear now. Thanks for the quick feedback, closing.
I'm not sure if it's stated somewhere in the docs tho.

kernc:
It's briefly mentioned [here](https://kernc.github.io/backtesting.py/doc/examples/Quick%20Start%20User%20Guide.html#Strategy). Docs improvements certainly welcome. :sweat_smile:

Issue #3, opened 2019-04-29
Option to limit order size/exposure


adrianmiu:
Is there a way to place orders that are limited in size? 

I'm working on a Forex strategy where I want to specify the size of the order based on the SL/TP levels provided by a prediction model. Your library mentions that the Strategy.buy/sell methods use "all available funds" which cannot be used in this case.

Thank you!

kernc:
Would you like to try to make a PR for it? I think API-wise, something like this might work?
```py
Strategy.buy(..., size: float = 1)  # percentage of cash
```

adrianmiu:
I would if I were a python programmer. Unfortunately I'm just hacking around... started playing with python a month ago. With that being said, it is an interesting challenge but I won't make any promise :)

Pk13055:
I have monkey patched the code to make it work with variable position sizing. I will clean it up and submit a PR.

kernc:
@Pk13055 If you find time, I'd appreciate it! I know I've been looking into it briefly and it seemed harder than I was willing to handle at the time. :sweat_smile:

packandsell:
Gents, I'm also working with FOREX markets and also with leverage of 1:100 and up. Usually my position size is estimated in so called Lots (100k USD), the money management would be 0.1 Lots per 1000 USD balance. Could you please advise on how to simulate this?

hotsun:
I updated:

    @property
    def size(self):
        """Position size in units of asset. Negative if position is short."""
        return int(self._broker._position)

and it looks ok for me.

Issue #10, opened 2019-07-13
How to use backtesting.py for forex trading


sylver:
### Expected Behavior

Having an example in the documentation on how use this library for Forex trading with lot size, position size, leverage, margin call, multiple positions opening, etc

### Actual Behavior

I can't figure out how/if this is possible with this library even if you mention forex and have a sample set of EURUSD.

The methods `buy` and `sell` of `Strategy` mention that they "close any current position and use all available funds" which seems a bit extreme and inflexible when money management recommends to use only 1% of your account on each trade.
Maybe I didn't understand what you mean here but I don't see the point to use all or none of the account and not give a strategy a way to set the account's amount/percent we want per trade and to open several trades on the same pair, moreover for Forex with lots.

If not already coded, would love to help improve that.

Disclaimer: I'm new to trading and forex trading, maybe there's something I didn't understand on this part.

Awesome library btw

packandsell:
+1 here. at least it would be helpful to be able to 
1) simulate 1:100 leverage (there's an example with 1:30, but no formula on how to setup any leverage)
2) fixed and variable (on demand) position size in trading assuming lots (100k USD is one lot roughly)

baptisteArnaud:
https://kernc.github.io/backtesting.py/doc/backtesting/backtesting.html

in the init function you can specify a margin option
> margin is the required margin (ratio) of a leveraged account. No difference is made between initial and maintenance margins. To run the backtest using e.g. 50:1 leverge your broker allows, set margin to 0.02

but for variable position size I'm looking for this option too !

jbobos:
Hi @kernc ,

I am wondering if the best way to add real forex trading support is to create a subclass of _Broker and implement all method with my real forex trading APIs. Any suggestion is appreciated. BTW, I'm using OANDA v20.

Thanks in advance.

packandsell:
Dear @kernc , is there a possibility to offer you a small job of integrating your lib into my project? I've tried doing it other developer following the guidelines in the questions asked here, but eventually it stops working either in one way or another due to the way signals are generated in my project. Of course, that would be paid. Really it should take no more than one for you to get it done since your know the ins and outs of this projects. If you would help on this one, please drop me a line on  the email edward.samokhvalov@gmail.com or leave your contact. Thanks.

kernc:
Sorry for not chipping in sooner. I guess order sizing and what came with it required a more serious overhaul.

[...]

PR #24, opened 2019-11-04
Feat/order size

kongyujian:
Added order sizing feature. Usage is as per discussion in #3 

Opted to use absolute position size instead of a percentage based off holding cash.

Ran with the code on the example, with a tweak on sizing:

```python
from backtesting.lib import crossover

from backtesting.test import SMA, GOOG


class SmaCross(Strategy):
    n1 = 10
    n2 = 30

    def init(self):
        self.sma1 = self.I(SMA, self.data.Close, self.n1)
        self.sma2 = self.I(SMA, self.data.Close, self.n2)

    def next(self):
        if crossover(self.sma1, self.sma2):
            self.buy(size=1000)
        elif crossover(self.sma2, self.sma1):
            self.sell(size=1000)


bt = Backtest(GOOG, SmaCross, cash=10000, commission=.002)

output = bt.run()
bt.plot()
```

On a quick eyeball, it seems to work.

Did write one simple test. Let me know if there needs be further coverage.

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -193,7 +193,7 @@ class Strategy(metaclass=ABCMeta):
         one at take-profit price (`tp`; limit order).
 
         If `price` is not set, market price is assumed.
-        
+
         If `size` is not set, entire holding cash is assumed.
         """
         self._broker.buy(price and float(price),
@@ -211,7 +211,7 @@ class Strategy(metaclass=ABCMeta):
         one at take-profit price (`tp`; limit order).
 
         If `price` is not set, market price is assumed.
-        
+
         If `size` is not set, entire holding cash is assumed.
         """
         self._broker.sell(price and float(price),
@@ -469,11 +469,11 @@ class _Broker:
     def __repr__(self):
         return '<Broker: {:.0f}{:+.1f}>'.format(self._cash, self.position.pl)
 
-    def buy(self, price=None, sl=None, tp=None, size:float=None):
+    def buy(self, price=None, sl=None, tp=None, size: float = None):
         assert (sl or -np.inf) <= (price or self.last_close) <= (tp or np.inf), "For long orders should be: SL ({}) < BUY PRICE ({}) < TP ({})".format(sl, price or self.last_close, tp)  # noqa: E501
         self.orders._update(price, sl, tp, size)
 
-    def sell(self, price=None, sl=None, tp=None, size:float=None):
+    def sell(self, price=None, sl=None, tp=None, size: float = None):
         assert (tp or -np.inf) <= (price or self.last_close) <= (sl or np.inf), "For short orders should be: TP ({}) < BUY PRICE ({}) < SL ({})".format(tp, price or self.last_close, sl)  # noqa: E501
         self.orders._update(price, sl, tp, size, is_long=False)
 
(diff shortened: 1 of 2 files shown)


backtesting/backtesting.py
```diff
@@ -493,7 +503,8 @@ def _open_position(self, price, is_long):
 
         i, price = self._get_market_price(price)
 
-        position = float(self._cash * self._leverage / (price * (1 + self._commission)))
+        size = self._cash if self.orders._size is None else self.orders._size
+        position = float(size * self._leverage / (price * (1 + self._commission)))
```
kernc: Is this `position` computation correct? You effectively replace former `cash` with `size`, but size actually corresponds to `cash / price`?

I think when size is given, position might better be `float(size) if enough_cash else error`?

kongyujian: The intention for the `size` variable being computed as such is because:

1. If size is not specified when performing an order, it will default to the default behaviour of using entire holding cash amount.
2. I did not notice there was a `margin` option to define a leveraged trade. (Only found out about it after reading: #10 ) Was initially under the assumption that all trades were on a 1:1 leverage. Under my assumption, i believed that specifying the `margin` option wasn't needed in order to define a "leveraged trade". In fact, by specifying an order size that is larger than your holding cash amount would technically mean you enter a trade on cross margin. (which was why I did not implement a check for `if enough_cash`). Reason for this assumption is largely influenced by Trading View's pine script, and how "leveraged trades" would be entered under their backtesting tools.

So my thoughts are as follows, and would need some clarification:

1. Does the library actually calculate liquidations. If it does, the `margin` option would come in handy for that. A quick look into the code base shows how `margin` is used to calculate `leverage` amount. But `margin` and `leverage` are simply being used in reporting functions and likely no where else?
2. Would it make sense to have margin be calculated dynamically if order size was specified, and larger than than holding cash. In that case, we could dynamically calculate a cross margin on init instead.

Would love to take a stab at pyramiding too, there are indeed many use cases for that. Let me know your thoughts. I might be working on a path different than the philosophy of this library. :)

kernc: > Does the library actually calculate liquidations.

It simulates them.

https://github.com/kernc/backtesting.py/blob/b3cfaff2956d19b3c4ea6329bb915de205fd7032/backtesting/backtesting.py#L561-L570

`margin=` therefore represents initial and maintenance margin of the trades. It's used only in stats and plots so the interested user gets a rough idea on how leveraged their strategy is allowed to be to still remain afloat.

kernc: In our case, `position` is actually equal to `size` when the user requests it. Thus, we only need to deduct the commission and then confirm we have (leveraged) cash to cover it all.

kernc: I'm really sorry for the long time I took to respond. I'm completely elsewhere, and this seemed like a much harder issue to grasp than it really was.

kernc: What do you think? :sweat_smile:

javamon1174:
An error message is displayed during this operation.

`self.sell(tp=7200)`

An error occurs when only a target price is specified for a sell order.

Conditional error in broker class

Please solve.

drajtrade:
Does this change work properly?

P-a-v-e:
@kernc I am just starting out with backtesting and it seems like the order sizing isn't in place. I am running backtesting version 0.1.7 however the changes above seem no to be commited.

kernc:
@P-a-v-e PR https://github.com/kernc/backtesting.py/pull/47 hasn't been merged yet. Welcome to try it out!

zachrdlapetr:
@kernc when you expect to release the 0.2.0 version with this feature?

PR #33, opened 2019-12-31
Travis CI: sudo: is deprecated, Xenial is default

cclauss:
Try Python 3.8


kernc:
Examples-running unit test has been failing since a couple of weeks and I guess needs some investigating.

kernc:
I think this PR is obsolete. Apparently I bumped the versions some time ago. Thanks anyway! :sweat_smile:

Issue #35, opened 2020-01-03
Error when plotting a large amount of data


thomasd3:
I have a very simple test:

```
df=load_dataframe(r’./historical.xlsx')
df = df.set_index('date')

# the backtest is a strategy that does nothing for now
bt = Backtest(df, empty_strategy, cash=1, commission=.002)
bt.run()

# and the display simply
bt.plot()

```

with all the data, bt.plot() will fail with _Javascript Error: Maximum call stack size exceeded._

if I add:

```
df = df[0:50000]
```

then the display works fine.

I have about 1.3M data points (and need them); is there a way to resample the data but ONLY for the display?

thomasd3:
I just realized that even with 50k data, it fails, but 20k is fine.

kernc:
> is there a way to resample the data but ONLY for the display?

Not presently. I think the exact amount might depend on available RAM and web browser configuration. Would certainly welcome automagically downsampling to a reasonable amount of candles for plot. :+1:

Issue #29, opened 2019-11-22
Track the actions of the trade strategy


Gnommer:
if we say are defining the next in the strategy class like buy and sell actions how can we track this in the _trade_data object so we can track easily when the order was placed and when the broker executed the buy sell and hold to present in the plots ?

- Backtesting version: 0.1.2

Gnommer:
@kernc any suggestions on the above issue ?

kernc:
`results._trade_data`, as mentioned in https://github.com/kernc/backtesting.py/issues/4, is only available _after_ the simulation run. For the time being, if you need to query state _during_ the simulation, you should look into the properties of [`Position`](https://kernc.github.io/backtesting.py/doc/backtesting/backtesting.html#backtesting.backtesting.Position) (`Strategy.position`) and track state changes manually yourself.

Benouare:
Just for help : 
```
final = pd.DataFrame()
final = final.append(trades[trades[["Entry Price"]].notnull().all(1)])
final = final.append(trades[trades[["Exit Price"]].notnull().all(1)])
final = final.sort_index()
```
It looks like this just shows the buy/close position dates.

Issue #28, opened 2019-11-19
New Sell order followed by Stop Loss  - double `_close_position` bug


Voyz:
It's a rare situation - you need both stop loss and new order to happen in the same time step - but when happens it is such a pain to nail down what's really wrong, since the framework won't break. Instead, all the data following will be slightly wrong. Started debugging this when trying to understand why `ohlc_trades` plot would draw two trades going on at the same time, if clearly existing positions need to be closed before opening new ones. That "double trade" plotting is result of that misaligned data due to the erroneous open position and instant stop loss.

![image](https://user-images.githubusercontent.com/15874847/69183969-3b474780-0b14-11ea-9e85-f17ca0c205bb.png)

### Expected Behavior
Close the original position, ignore the stop loss and create a new order

### Actual Behavior
Close the original position, create a new order, and instantly close it due to stop loss

### Steps to Reproduce

1. Create a strategy that will produce a new sell order the same time step in which a stop loss threshold is reached.

### Additional info
A couple of fix suggestions:

1. Update `open, high, low` once `self._open_position(entry, is_long)` is called ([backtesting.py line 539](https://github.com/kernc/backtesting.py/blob/master/backtesting/backtesting.py#L539))

2. 
[backtesting.py line 532](https://github.com/kernc/backtesting.py/blob/master/backtesting/backtesting.py#L532)
```python
if entry or orders._close:
    self._close_position()
    orders._close = False

################ THIS
# First make the entry order, if hit
if entry:
    if entry is _MARKET_PRICE or high > orders._entry > low:
        self._open_position(entry, is_long)
################ THIS

# Check if stop-loss threshold was hit
if sl and self._position:
    price = (sl if low <= sl <= high else              # hit
             open if (is_long and open < sl or         # gapped hit
                      not is_long and open > sl) else
             None)                                     # not hit
    if price is not None:
        self._close_position(price)
        self.orders.cancel()

# Check if take-profit threshold was hit
if tp and self._position:
    price = (tp if low < tp < high else
             open if (is_long and open > tp or
                      not is_long and open > sl) else
             None)
    if price is not None:
        self._close_position(price)
        self.orders.cancel()
   
###########################################
#### SHOULD PROBABLY BE MOVED HERE #######
###########################################

```

kernc:
Apologies for the inconvenience. :sweat_smile: You are welcome to propose a more ready fix as part of a pull request. 

[...]

Issue #4, opened 2019-05-22
Way to access equity values, individual trade returns, ...


fabio11c:
Is there a way to extract the returns of the strategy and not only have a plot?

kernc:
```py
bt = Backtest(...)
result = bt.run()

result._trade_data  # type: pd.DataFrame
```
In `_trade_data`, you can find columns such as changing Equity values, P/L of individual trades, ...
https://github.com/kernc/backtesting.py/blob/a7d1d4f3fa658290182dfa2fdbf69bffe6a8f018/backtesting/backtesting.py#L849-L865

Note, this is currently private API, so it may change.

fabio11c:
Ok, thanks...what do you mean with a "private API"?

fabio11c:
`result._trade_data` is not an atribute of `bt.run()`. Im still having the issue.

kernc:
`bt.run()._trade_data`

It should be. It's the way plotting functions access raw trading data:

https://github.com/kernc/backtesting.py/blob/a7d1d4f3fa658290182dfa2fdbf69bffe6a8f018/backtesting/backtesting.py#L911
https://github.com/kernc/backtesting.py/blob/a7d1d4f3fa658290182dfa2fdbf69bffe6a8f018/backtesting/_plotting.py#L80

kernc:
> what do you mean with a "private API"?

It's undocumented API, without reliability (change) guarantees, not meant for public consumption. It's why the attribute is prefixed with an underscore, following [a common Python convention](https://docs.python.org/3/tutorial/classes.html#private-variables).

You are welcome to use it in testing, though. Maybe even let me know how you are using it and whether you need/expect something else from/like it too. :+1:

packandsell:
Hello, everyone. Is there a way to have final PnL in pips or calculate it somehow?

kernc:
<sup>Hello. :smiley: This is not a forum, and even the forums are clearer when restrained to one topic per thread.</sup> 

This might do it for you, within `Strategy.next()`:
```python
pl_pips = self.position.pl / self.data.pip
```
https://kernc.github.io/backtesting.py/doc/backtesting/backtesting.html#backtesting.backtesting.Strategy.data

-----

Note to self: `data` should be better documented.

kernc:
Oh, you mean final `result['P/L']`. I guess you can just divide it by pip size? If you don't know what that is, backtesting.py uses the following heuristic formula to determine it from Close values:
[...]

PR #47, opened 2020-03-09
Order/Trade/Position API

kernc:
refs https://github.com/kernc/backtesting.py/issues/8
fixes https://github.com/kernc/backtesting.py/issues/28
closes https://github.com/kernc/backtesting.py/issues/77
closes https://github.com/kernc/backtesting.py/issues/83
closes https://github.com/kernc/backtesting.py/issues/92

**New features:**
* Add order sizing, pyramiding, order types (stop-limit order) (fixes https://github.com/kernc/backtesting.py/issues/3, fixes https://github.com/kernc/backtesting.py/issues/10, closes https://github.com/kernc/backtesting.py/pull/24, closes https://github.com/kernc/backtesting.py/issues/76, closes https://github.com/kernc/backtesting.py/issues/106),
* `Strategy.orders` list of active orders,
* `Strategy.trades` list of active trades,
* `Strategy.closed_trades` list of settled trades,
* `Order` class, representing an active order,
* `Trade` class, representing an active or already settled trade,
* auto-downsample plots when too much data (fixes https://github.com/kernc/backtesting.py/issues/35),
* New `stats` keys (fixes https://github.com/kernc/backtesting.py/issues/4, fixes https://github.com/kernc/backtesting.py/issues/29):
  * `_equity_curve`, equity, drawdown, and drawdown duration,
  * `_trades`, DataFrame of executed trades.

**Breakages:**
* Need to call explicit `Strategy.position.close()` between subsequent `Strategy.buy/sell()` calls to approximate previous behavior.
* `Strategy.buy(price=)` → `Strategy.buy(limit=)`
* `Strategy.buy/sell()` only takes keyword parameters.
* `Position.open_price` → `Trade.entry_price`
* `Position.open_time` → `Trade.entry_time`
* `Strategy.orders` → `Strategy.orders[i]`
* `Backtest.plot(omit_missing=)` is gone.
* Revised `backtesting.lib.SignalStrategy`.


qacollective:
Awesome framework! This looks like a great PR too.

If I wanted to disable the NFA compliant FIFO functionality, would that be as simple as commenting out lines 824 to 842 in backtesting.py?

This rule only applies to brokers in the USA, so perhaps this could even be a configuration option?

backtesting/backtesting.py
```diff
@@ -454,128 +653,248 @@ def __init__(self, *, data, cash, commission, margin, trade_on_close, length):
...
+                # Not enough cash/margin even for a single unit
+                if not size:
+                    continue
+            assert size == round(size)
+            need_size = int(size)
+
+            # Fill position by FIFO closing/reducing existing opposite-facing trades.
+            # Existing trades are closed at unadjusted price, because the adjustment
+            # was already made when buying.
+            for trade in list(self.trades):
```
kernc, line 827: > If I wanted to disable the NFA compliant FIFO functionality, would that be as simple as commenting out lines 824 to 842 in backtesting.py?

@qacollective Yes, I believe so.

I was under the impression the rule was enforced for all _retail_ traders. It's at least consistent with what I see on Oanda Europe. Maybe only applies to forex. We should definitely make this optional. :+1:

backtesting/backtesting.py
```diff
@@ -454,128 +653,248 @@ def __init__(self, *, data, cash, commission, margin, trade_on_close, length):
...
     @property
     def equity(self):
-        return self._cash + self.position.pl
+        return self._cash + sum(trade.pl for trade in self.trades)
+
+    @property
+    def margin_available(self):
+        # From https://github.com/QuantConnect/Lean/pull/3768
+        margin_used = sum(trade.value / self._leverage for trade in self.trades)
+        return self.equity - margin_used
```
kernc, line 719: @qacollective Thanks for the review. If you consider trying the PR further, I seem to have misplaced the offending failing/raising test at the moment, but be advised there might be a bug somewhere in/with `_Broker.margin_available()` calculation. :sweat_smile:

kernc: And not necessarily the same, but related: https://github.com/kernc/backtesting.py/issues/77.

backtesting/backtesting.py
qacollective: @kernc I'm only just learning of this too, but from my current understanding it appears that hedging Forex is illegal in the USA (the land of the 'free' :roll_eyes:) but legal almost everywhere else (https://www.forexbonuses.org/trading-methods/hedging/).

There are even broker review websites which help you find brokers who don't comply with FIFO: https://www.benzinga.com/money/hedging-forex/ :laughing:

backtesting/backtesting.py
qacollective: @kernc No worries!!  I'm a bit new to GitHub as well, so should have submitted a proper review with suggested code change which I see is now possible from here.  

FYI I really liked the look / feel of backtesting.py when I first saw it, but the fact that it allocated 100% of the funds and didn't seem to be able to hold multiple concurrent orders turning into multiple simultaneous positions rendered it a non-candidate for my use until I stumbled across this PR - which on my current understanding, changes all of those things in one big hit :ok_hand:

At the moment the only thing that I'm not new to here is Python.  So I'm quite tentative.  I'm in the early days of getting to understand backtesting.py so I can run it over some 100% fake but mathematically pure and simple data and strategy to test that it works to my understanding and requirements.  I will let you know if I find anything that doesn't look right! :thinking:

kernc: > so I can run it over some 100% fake but mathematically pure and simple data and strategy to test that it works to my understanding and requirements. I will let you know if I find anything that doesn't look right! thinking

Even much less than that would be seriously awesome and very appreciated. :smiling_face_with_three_hearts:

sdmovie:
one bug found:
I set cash to 10000 and strategy for NQmain whcih is ~9200~9700 price range in Feb.Strategy check indicator and position - when sell indicator and position>=0,sell ; when buy inidicator and position<=0, buy. 
several normal unsuccessful trades executed then equity fall down to 9400$.
startegy try to place an sell order whenprice is 96xx$, it's failed as expected due to insufficient cash.Then there are 3 consequent sell requests in and all fail. However the failed order si not killed but resides in self.orders.
when NQmain price fall, one order executed but other 3 still in self.orders. When a buy order in ,the short position covered.But the order in queue automatically executed and make a mess.
in _broker -- _process_order:
```
# If we don't have enough liquidity to cover for the order, skip it
            if abs(need_size) * adjusted_price > self.margin_available * self._leverage:
                continue
```
should the order be removed as it fails fro insufficient cash?

sdmovie:
for the previous margin_available calculation prohibit buyback of short position ( and I think vice versa for sell long position). The opposite order ( buy for short position , sell for long position) should  not be constrained by margin _available. Opposite orders bypass margin_available check, only open (long or short) orders do. 
thus avoid dealing with the "frozen cash" concept which might be more complex.

kernc:
> ```py
>     if np.sign(self.position.size) + np.sign(order.size) == 0:
>         size=np.sign(size)*round(abs(size)*abs(self.position.size))
> ```
`size=` computation overrides the size/portion set by the user (could be e.g. 
```py
self.buy(size=.1)
# ...
self.sell(size=.05)  # order.size = 0.5
```
This also likely wouldn't work for hedging traders that run long and short trades in parallel (see https://github.com/kernc/backtesting.py/pull/47#discussion_r430470327).

I think a better approach would be to fix `margin_available()` to take into account order direction.

sdmovie:
would it be more clear to distiguish order as open(long or short) orders and close orders(long or short)? an open order leverages available cash(percentage) , and close orders count on holding positions( percentage)?

backtesting/backtesting.py
```diff
@@ -454,128 +653,248 @@ def __init__(self, *, data, cash, commission, margin, trade_on_close, length):
...
+                    assert order.size != -trade.size or trade not in self.trades
+                if order in (trade._sl_order,
+                             trade._tp_order):
+                    assert order.size == -trade.size
+                    assert order not in self.orders  # Removed when trade was closed
+                else:
+                    # It's a trade.close() order, now done
+                    assert abs(trade.size) >= abs(order.size) >= 1
+                    self.orders.remove(order)
+                continue
```
kernc, line 804: > https://github.com/kernc/backtesting.py/pull/47#issuecomment-635372131

@sdmovie Explicit closing orders (SL/TP orders and orders made through `self.trades[i].close(portion)` or `self.position.close(portion)`) are already exempt from margin computation.

sdmovie: Great! that position close already fullfill my use case. Not necessary to struggle on buy/sell margin calculation - for me

sdmovie: how does resample work to comply various open hours (start time unalign and segmented hours)? 
US stock 9:30-16:00 - should be 9:30-10:30,10:30-11:30,...15:00-16:00
US future 17:00 -16:15 should be 17:00-18:00,... 16:00-16:15
HK stock: 9:30-12:00,13:00-16:00  -- should be 9:30-10:30,10:30-11:30,11:30-12:00,13:00-14:00 .. 15:00-16:00

I have s draft proposal:

sdmovie:
how does resample work to comply various open hours (start time unalign and segmented hours)? 
US stock 9:30-16:00 - should be 9:30-10:30,10:30-11:30,...15:00-16:00
US future 17:00 -16:15 should be 17:00-18:00,... 16:00-16:15
HK stock: 9:30-12:00,13:00-16:00  -- should be 9:30-10:30,10:30-11:30,11:30-12:00,13:00-14:00 .. 15:00-16:00

sdmovie:
I read through bacltrader's resample, cannot understand the code - it's too complicate.Not sure how it resample above cases if it did work with various open hours and boundary

sdmovie:
one more change is to enable strategy choose brokers ,thus can split backtest for future/us stock /forex etc to standalone brokers, instead of mixing up everrthing in one process/function, expect to be easier for implementation

sdmovie:
@kernc hi, are you going to use dataframe for trades and maybe orders,positions?As I am building realtrading , would like to align with you to get minimum difference from the backtest
```
New stats keys (fixes #4, fixes #29):
_equity_curve, equity, drawdown, and drawdown duration,
_trades, DataFrame of executed trades.
```

sdmovie:
personally I vote for dataframes(1 order per row) vs. list of class objects
- can easily compare last and this order status and deal only the changed ones(rows)
- orders/trades/positions from real broker are often dataframes or easy convertable to df

kernc:
> how does resample work to comply various open hours (start time unalign and segmented hours)?
> US stock 9:30-16:00 - should be 9:30-10:30,10:30-11:30,...15:00-16:00
> US future 17:00 -16:15 should be 17:00-18:00,... 16:00-16:15
> HK stock: 9:30-12:00,13:00-16:00 -- should be 9:30-10:30,10:30-11:30,11:30-12:00,13:00-14:00 .. 15:00-16:00

The only place where Backtesting.py does resampling is in `Backtest.plot(superimpose=...)`, i.e. when plotting superimposed OHLC overlay.

https://github.com/kernc/backtesting.py/blob/b460092c5e832706596ac6652eb1762b22044a95/backtesting/_plotting.py#L367

Everywhere else we just use whatever bars are available in user-provided `data` and interpret them as bars in sequence.

kernc:
> what if adding optional data1,data2,data3 to Strategy, cause only strategy might use extra timeframe data (also Backtest class to pass data to strategy). brokers are not impacted.

Not in favor of extra data parameters. Resampling and data shaping is the domain of and easy enough in [pandas](pandas.pydata.org/). Note, `data` passed into Backtest can already contain any extra data columns (https://github.com/kernc/backtesting.py/issues/64), and there's a recipe published for [strategy testing on multiple timeframes simultaneously](https://kernc.github.io/backtesting.py/doc/examples/Multiple%20Time%20Frames.html) ([`backtesting.lib.resample_apply()`](https://kernc.github.io/backtesting.py/doc/backtesting/lib.html#backtesting.lib.resample_apply) _also_ doesn't care about workday time offsets of different exchanges but just passes on the provided [pandas' datetime offset string](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#dateoffset-objects) (or object, e.g. `pd.offsets.CustomBusinessHour`)).

kernc:
> @kernc hi, are you going to use dataframe for trades and maybe orders,positions?

If you mean the new mentioned dataframes `stats['_equity_curve']` and `stats['_trades']`, then yes, they are intended as public API despite the underscore. Positions can be inferred from the `_trades` data. The internal _orders_ I'd prefer to keep a list (sequence) of our `backtesting.Order` or `Order`-like objects.

kernc:
> orders/trades/positions from real broker are often dataframes or easy convertable to df

@sdmovie And dataframes, compared to simple dicts, namedtuples, or thin dataclasses are also _painfully slow_. I was considering it for a time, but the few advantages are minor. :sweat_smile:

arunavo4:
@kernc There is an issue here 

```python
mean_return = returns.prod()**(1/(len(returns) or np.nan))
```
I don't understand how this can be mean_return. I found it cause it was throwing errors.
```
/backtesting/backtesting.py:1290: RuntimeWarning: invalid value encountered in double_scalars
  mean_return = returns.prod()**(1/(len(returns) or np.nan))
```

I changed it to this for now.
```python
mean_return = returns.mean()
```

kernc:
> ```
> RuntimeWarning: invalid value encountered in double_scalars
> ```

@arunavo4 I wonder, what are the values of `returns` when this error is thrown?

arunavo4:
> > ```
> > RuntimeWarning: invalid value encountered in double_scalars
> > ```
> 
> @arunavo4 I wonder, what are the values of `returns` when this error is thrown?

```
0    -0.061939
1     0.013994
2    -0.032009
3    -0.045547
4    -0.100880
        ...   
88    0.055558
89   -0.037364
90    0.012820
91    0.214068
92    0.002739
```

kernc:
Can you `print(results.tolist())`? Something I can fully reproduce the error with. :smiley:

arunavo4:
> Can you `print(results.tolist())`? Something I can fully reproduce the error with.

Just Running the Below Strategy from your examples you can reproduce the error
```python
class SmaCross(Strategy):
    def init(self):
        close = self.data.Close
        self.ma1 = self.I(SMA, close, 10)
        self.ma2 = self.I(SMA, close, 20)

    def next(self):
        if crossover(self.ma1, self.ma2):
            self.position.close()
            self.buy()
        elif crossover(self.ma2, self.ma1):
            self.position.close()
            self.sell()

bt = Backtest(GOOG, SmaCross,
              cash=10000, commission=.002)
stats = bt.run()
```
@kernc Here are the returns to reproduce
```python
>>> print(returns.to_list())
[-0.06193928503083557, 0.013993895756747587, -0.03200907309123746, -0.04554720912935384, -0.10087973123689231, -0.020541370548910365, 0.02686905108717197, 0.5711930956177456, 0.03074631753779944, 0.03853295891689301, -0.1662989845236693, 0.18885481895167744, -0.013032124855772098, 0.017866434788202046, -0.007940553133350958, 0.01465838481500037, 0.11478128671135368, -0.12446769225016019, -0.15670030218828845, 0.013184710001359567, 0.01666104601701701, -0.0061398801365677436, 0.03188570127254031, 0.016167514868408395, -0.0007378855956048369, 0.2677667938326673, -0.033731154885610515, -0.03128176470871602, 0.03756872440307568, 0.003601684560982532, -0.031130156913634188, 0.057984094350385984, -0.014791859297223153, 0.22001734560470165, -0.10249247363037206, 0.0012599344783359445, 0.3544694534466708, 0.2191634954952737, -0.006767081039364076, 0.01966893783580681, 0.09525367856231082, -0.07530252026765616, 0.24029005323046304, -0.009617781507963441, 0.1376766661435137, 0.0925317068135485, 0.04223125811957884, -0.06546088418864149, -0.12387572084176401, 0.032844120607389504, -0.0035680101097095296, 0.1289110091505301, -0.07046597340618233, -0.030787563935487428, -0.06693160081382499, 0.29701336357308494, -0.014165590612924772, 0.03967218312865106, 0.08300775886893619, 0.03445043614428611, -0.01030018746161021, -0.041600318016278504, 0.09082895960554249, -0.06464372662214812, 0.00320665498094419, 0.013027195938903446, -0.027655242103755695, 0.21644407819427292, -0.015449702824452238, 0.028557711262134466, -0.025996346211244026, -0.04319998335889008, 0.00935582401882784, -0.08229073051199465, 0.017621150717139544, -0.005505571039324586, -0.0024178027746595276, 0.051242107210665244, 0.051507125548257204, -0.04246831891354663, -0.14024263099111822, -0.003406517081832927, -0.07407109297733139, 0.030063407900160044, 0.04403261975674477, 0.013262744224080913, -0.02414560312946601, 0.018353768503395873, 0.055558418203095816, -0.037363954813400935, 0.012820449977274828, 0.21406813104364875, 0.002739172610932372]
```

kernc:
@arunavo4 Thanks. I fixed the geometric mean of negative numbers by adding +1 to contain all returns around 1 instead of 0. Seems to work and is the correct approach as returns of 1 (× 1.0 × 1.0 × ...) maintain the identity.
https://github.com/kernc/backtesting.py/blob/e56f75830178f5fd49bb9dbc736c4aebce3b499c/backtesting/backtesting.py#L1288

kernc:
@sdmovie I fixed the [stale orders](https://github.com/kernc/backtesting.py/pull/47/#issuecomment-635208951) issue, but the [`margin_available` issue](https://github.com/kernc/backtesting.py/pull/47/#issuecomment-635228932) (i.e. #77) I ended up disagreeing with. At that point in time that is the remaining available margin. If one needs to close existing trades before placing new ones, one can call `self.position.close()` or run `Backtest(..., exclusive_orders=True)` which does it automatically.

kernc:
@qacollective I added `Backtest(..., hedging=True)` switch that disables NFA FIFO behavior.

kernc:
Anybody happen to have an idea [why this fails](https://travis-ci.org/github/kernc/backtesting.py/jobs/707685491#L250-L260) on Python 3.5? :confused: 
```text
Traceback (most recent call last):
  File "/opt/python/3.5.6/lib/python3.5/runpy.py", line 183, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/opt/python/3.5.6/lib/python3.5/runpy.py", line 109, in _get_module_details
    __import__(pkg_name)
  File "/home/travis/build/kernc/backtesting.py/backtesting/__init__.py", line 52, in <module>
    from .backtesting import Backtest, Strategy  # noqa: F401
  File "/home/travis/build/kernc/backtesting.py/backtesting/backtesting.py", line 940
    ):
    ^
SyntaxError: invalid syntax
```
https://github.com/kernc/backtesting.py/blob/b123fba66bbe6d6f8441950e3620a30cd814d341/backtesting/backtesting.py#L925-L935

qacollective:
@kernc 

Trailing commas in function declarations were only allowed from Python version 3.6, I think?  Or it is the '*'?

https://bugs.python.org/issue9232

TL;DR !

kernc:
As a simple performance benchmark, 0.1.8 CI tests take [2min 40sec](https://travis-ci.org/github/kernc/backtesting.py/builds/708148049) versus this PR [3min 1sec](https://travis-ci.org/github/kernc/backtesting.py/builds/708171062). It's the same ballpark figure, so its fine.

kernc:
Much thanks to @qacollective, @sdmovie, and @arunavo4 for reviews and help!

Benouare:
Hi, 
About the first post of this issue : "Position.open_time → Trade.entry_time" 
You should change this line, the entry_time doesnt not exist and you have to do something like this
`self.data.index[self.trades[-1].entry_bar]`
To get the entry time. 

It confused me at the begining. 
Cheers
FYI : https://github.com/kernc/backtesting.py/issues/117

PR #85, opened 2020-06-08
Add profit factor to stats

### LINKED ISSUE #84: Profit factor statistic
Hey, this is one great backtesting framework. 
I'm surprised that there is no profit factor (gross profit/gross loss) in backtest stats. 
Why is that? Is that something you could add?

TomaszLakota:
profit factor = gross profit / gross loss

Fixes #84


kernc:
That's perfect, thank you!

PR #110, opened 2020-07-17
typo fix: 'Windos'

tankorsmash:
Tiny typo fix in the docs. I swear there was another one, like 'com' instead of 'come' somewhere, but I couldn't find it.


tankorsmash:
Found another one 

https://github.com/kernc/backtesting.py/blob/64bcda2ffea565327ecdeb3818fa10183c16105c/backtesting/backtesting.py#L1230

kernc:
Thanks. These are intentional. :sweat_smile: :heart:

PR #111, opened 2020-07-17
fix typo in 'commision'

tankorsmash:
Tiny typo fix in the docs. I swear there was another one, like 'com' instead of 'come' somewhere, but I couldn't find it.


kernc:
Thanks! I quickly scoured the source for strings containing 'com', and nothing further out of the ordinary popped up. Welcome it later. :+1:

Issue #150, opened 2020-09-30
Clarification suggestion for stats output "Buy" & Hold label in output


blockforcecapital:
First off, Thank you for the wonderful work.  This is very useful and I am sure will be appreciated by many.  I've dabbled with a few packages for backtesting in python and this is by far my favorite.

The following is just my opinion, and certainly not critical.  You commented it in your source code but for beginners, it might be confusing.  Here is the line I am referring to: https://github.com/kernc/backtesting.py/blob/2bb8f29c393fec030cfb247b16eea2aec2304f45/backtesting/backtesting.py#L1383

### Expected Behavior
In a declining market, the stats line "Buy & Hold Return [%]" would normally show a negative return.
### Actual Behavior
Instead, the result is calculated based on the absolute value calculation. This could be confusing to some.  Perhaps add a new line "Short & Hold Return [%]" or change the line to "Best of (Long or Short) & Hold Return [%]"

### Steps to Reproduce

1. Chose a timeframe where your data series loses value
2. stats = bt.run() then print (stats)
3. Buy & Hold Return [%] will show a positive number.

### Additional info

- Backtesting version: 0.2.0

kernc:
Thanks for the warm intro. It does look promising at first glance, but it still has many bugs to iron out.

It has indeed proven confusing a couple of times now, see e.g. https://github.com/kernc/backtesting.py/issues/140, https://github.com/kernc/backtesting.py/issues/36.

I see adding a new negative value key as somewhat pointless, but as an updated string, I'd certainly prefer something shorter, like `"Buy/Sell & Hold Return [%]"` or `"Buy (Sell) & Hold Return [%]"`. You think that would clarify it? Is there another similar expression that encompasses both meanings?

eervin123:
Hi there, I'm commenting on here from my other account. I didn't realize before I was logged in with @blockforcecapital before.

Thanks for responding, I'm sorry I didn't see #140 and #36 before, and I feel silly even wasting your time with such a trivial matter.  It took me a sec, but then I figured it out fairly quickly.  Given that, if you twisted my arm and I had to make a change, I would suggest you leave it as a negative number and keep `"Buy & Hold Return [%]"` I think that is how I am most accustomed to seeing it and instantly, my brain is able to process that a buy and hold investor would have lost money so, hopefully, my strategy would have a less negative return, or better yet, a positive return.

[...]

PR #152, opened 2020-10-04
Update stats calc for buy and hold to be long-only

eervin123:
Changed the "Buy & Hold" stats calculation to remove the absolute value function.  It will now reflect a negative number if the underlying asset has a negative return for the comparison period.

Fixes https://github.com/kernc/backtesting.py/issues/150.


kernc:
Lgtm, thanks!

PR #154, opened 2020-10-10
Model-based optimization with scikit-optimize

### LINKED ISSUE #69: Suggestion for Faster way to optimize vs GridSearch
Hey, 

First of all thank you for the hard work put into this, it's a really awesome project that I still learn how to use.
I'm not that proficient in using python so that is why I'll put my suggestion here. I have used python more for machine learning stuff and there the option of searching hyperparameters was also between grid and random search. However, I found that using bayesian optimization, code runs way faster because it add probabilities to the mix by building a surrogate probability model of the objective function. That being said, you can have a look maybe over this to find out more [here](https://github.com/fmfn/BayesianOptimization/blob/master/bayes_opt/bayesian_optimization.py)

My guess is that this can be applied to your _maximize_func_ in some way and computations should, in theory, run faster. 
I'll spend the next few days trying and playing around with this and if I come up with something, I'll let you know. 

Keep it up!

ttfreeman:
So just for initial review. I haven't added unit tests yet. If all is good, I can add unit tests. I probably need to spend more time until then on more refined understanding of strategies and other classes such as Order and Strategy .

Fixes #69.


backtesting/backtesting.py
```diff
@@ -1514,3 +1525,68 @@ def plot(self, *, results: pd.Series = None, filename=None, plot_width=None,
             reverse_indicators=reverse_indicators,
             show_legend=show_legend,
             open_browser=open_browser)
+
+    def optimize_skopt(self,
+                       maximize: Union[str, Callable[[pd.Series], float]] = 'SQN',
+                       max_tries: Union[int, float] = 100,
+                       **kwargs) -> Union[pd.Series, Tuple[pd.Series, pd.Series]]:
```
kernc, line 1532: No new API methods are planned. I suppose you'll have to find a way to integrate the changes within existing `Backtest.optimize()` API! Maybe simply checking if `param_combos > max_tries`
https://github.com/kernc/backtesting.py/blob/3045b64635fe9e5af51046b7bdf7cdfcbf375c1e/backtesting/backtesting.py#L1237-L1241
and then choose the skopt path.

Then again, `param_combos` can combinatorically explode, and no reason to compute it just to disregard it soon after ... :thinking: 

You're certainly welcome to defer (parts) to a `_optimize_skopt()` _private_ method/coroutine if you see fit.

kernc: > Then again, `param_combos` can combinatorically explode, and no reason to compute it just to not pay any attention to it later ... :thinking: 

I think we can avoid generating `param_combos` when not needed if we add an additional method argument:
```py
method: Literal['auto', 'full', 'random', 'skopt'] = 'auto'
```

backtesting/backtesting.py
```diff
@@ -1514,3 +1525,68 @@ def plot(self, *, results: pd.Series = None, filename=None, plot_width=None,
...
+            def maximize(stats: pd.Series, _key=maximize):
+                return stats[_key]
+
+        elif not callable(maximize):
+            raise TypeError('`maximize` must be str (a field of backtest.run() result '
+                            'Series) or a function that accepts result Series '
+                            'and returns a number; the higher the better')
+
+        dimensions = [Integer(name='n1', low=min(kwargs.get('n1')), high=max(kwargs.get('n1'))),
+                      Integer(name='n2', low=min(kwargs.get('n2')), high=max(kwargs.get('n2')))]
```
kernc, line 1576: Integer and min/max logic is fine for integer input, but we'll need to cover reals and categoricals also. I guess you can walk over `kwargs` values lists and wrap them as `np.array`. Then if `dtype.kind` is `'i'`, create an Integer dimension, if `'f'` (float), [a Real dimension, and a Categorical dimension](https://scikit-optimize.github.io/stable/modules/classes.html#module-skopt.space.space) in all other cases?

backtesting/backtesting.py
```diff
@@ -1514,3 +1525,68 @@ def plot(self, *, results: pd.Series = None, filename=None, plot_width=None,
...
+        res = forest_minimize(func=convert,
+                              dimensions=dimensions,
+                              n_calls=max_tries, base_estimator="ET",
+                              random_state=4)
+
+        self.run(**dict(zip(['n1', 'n2'], res.x)))
+        return_result = pd.Series([res.fun, res.x, res, self._results], index=[
+                                  "best_fitness", "best_parameters", "full_results", "stats_best"])
+
+        return return_result
```
kernc, line 1592: Return from `Backtest.optimize()` should remain best-run `stats: pd.Series`, but you are encouraged to add another method argument:
```py
return_optimization: bool = False
```
(akin to `return_heatmap=`) which additionally returns `scipy.optimize.OptimizeResult` as obtained from `forest_minimize()`. The user can then use `skopt.plots.plot_objective()` or the object's defined structure to draw further insights.

kernc (changes requested):
I looked over what seem to be the relevant changes. There's quite a lot of work to be done still! :smiley:

I think `Backtest.optimize()`, dispatching to a couple separate internal methods before tying in for a common result, is the best way to go. :+1: 

I somewhat mind all the unrelated/pedantry/style changes. In VSCode, you can stage for commit only certain hand-picked changed hunks (i.e. `git add --patch && git commit`) like so: https://stackoverflow.com/questions/34730585/how-can-i-commit-some-changes-to-a-file-but-not-others-in-vscode

backtesting/backtesting.py
ttfreeman: Sounds good. I will forge ahead and apply these and more in the next few days...

backtesting/backtesting.py
```diff
@@ -1163,6 +1174,8 @@ def run(self, **kwargs) -> pd.Series:
 
     def optimize(self,
                  maximize: Union[str, Callable[[pd.Series], float]] = 'SQN',
+                 method: Literal['auto', 'full', 'random', 'skopt'] = 'auto',
```
kernc, line 1177: The reason [CI breaks](https://travis-ci.org/github/kernc/backtesting.py/builds/736952138) is that [`Literal` was only included in Python 3.8](url), but we support 3.5+. You'll have to use a [type comment](https://www.python.org/dev/peps/pep-0484/#type-comments), or untype the line completely. :confused:

backtesting/backtesting.py
```diff
@@ -1221,101 +1236,172 @@ def maximize(stats: pd.Series, _key=maximize):
...
+            if return_heatmap:
+                return return_result["stats_best"], return_result["full_results"]
+            return return_result["stats_best"]
+
+        def switch_method(x):
+            return {
+                'auto': _auto(),
+                'full': 0,
+                'random': 0,
+                'skopt': _optimize_skopt(),
```
kernc, line 1398: Nice try. :smile: I actually meant `'auto'` as a default, to be a synonym for:
```py
'full' if len(param_combos) <= max_tries else 'random' if skopt is None else 'skopt'
```
The question is, why would we construct `param_combos` that can potentially explode and that we don't have intention of using. But if one is using `method='skopt'`, it suffices they provide just the min and max of ordinal values, so the parameter space might not be that large after all.

backtesting/backtesting.py
```diff
@@ -1221,101 +1236,172 @@ def maximize(stats: pd.Series, _key=maximize):
...
+                    index.append("'" + item + "'")
+                res = self.run(**dict(zip(index, params)))
+                return -res[maximize_key]
+
+            res = forest_minimize(func=convert,
+                                  dimensions=dimensions,
+                                  n_calls=max_tries, base_estimator="ET",
+                                  random_state=4)
+
+            self.run(**dict(zip(['n1', 'n2'], res.x)))
```
kernc, line 1385: This, working with `n1` and `n2`, doesn't produce the right `self._results` in the general case?

backtesting/backtesting.py
```diff
@@ -1221,101 +1236,172 @@ def maximize(stats: pd.Series, _key=maximize):
...
+            return_result = pd.Series([res.fun, res.x, res, self._results], index=[
+                "best_fitness", "best_parameters", "full_results", "stats_best"])
+
+            if return_heatmap:
+                return return_result["stats_best"], return_result["full_results"]
```
kernc, line 1390: We'll need to adhere to the same current API. Does `return_result['full_results']` (`= res`), in fact, produce a MultiIndexed pandas Series? I think for the optimized case, `return_heatmap` should return a heatmap (MultiIndexed pandas Series) of tried variable combinations.

Why not introduce a new argument `return_optimization: bool = False` that, additionally, packs in `res` at the end of the return tuple?

backtesting/backtesting.py
```diff
@@ -1221,101 +1236,172 @@ def maximize(stats: pd.Series, _key=maximize):
...
+            # @ use_named_args(dimensions=dimensions)
+            # def convert(n1, n2):
+            #     res = self.run(**dict(zip(['n1', 'n2'], [n1, n2])))
+            #     return -res[maximize_key]
+            @ use_named_args(dimensions=dimensions)
+            def convert(**params):
+                index = []
+                for item in params:
+                    index.append("'" + item + "'")
+                res = self.run(**dict(zip(index, params)))
```
kernc, line 1377: Given [what `use_named_args()` does to params](https://github.com/scikit-optimize/scikit-optimize/blob/de32b5fd2205a1e58526f3cacd0422a26d315d0f/skopt/utils.py#L800-L803), we can just do
```py
res = self.run(**params)
```
?

backtesting/backtesting.py
```diff
@@ -1221,101 +1236,172 @@ def maximize(stats: pd.Series, _key=maximize):
...
                             'and returns a number; the higher the better')
-
         if constraint is None:
 
             def constraint(_):
                 return True
 
         elif not callable(constraint):
-            raise TypeError("`constraint` must be a function that accepts a dict "
+            raise TypeError("constraint` must be a function that accepts a dict "
```
kernc, line 1245: While there are many changes here made by your opinionated editor (and which I'd very much like to eventually see reverted/cleaned up), this is probably not one of them. :smile:

backtesting/backtesting.py
kernc: And since we haven't checked params values with `constraint=` previously, we should check that they pass it now before `self.run()`.

backtesting/backtesting.py
```diff
@@ -1221,101 +1236,172 @@ def maximize(stats: pd.Series, _key=maximize):
...
+                                constraint=lambda p: p.sma1 < p.sma2)
+
+            .. TODO::
+                Improve multiprocessing with joblib.parallel .
+            """
+            dimensions = []
+            for key, value in kwargs.items():
+                args_array = np.array(value)
+
+                if args_array.dtype.kind == 'i':
```
kernc, line 1361: [`kind in 'ium'`](https://numpy.org/doc/stable/reference/generated/numpy.dtype.kind.html)

backtesting/backtesting.py
ttfreeman: Yes, sorry! I am still working on this part. The integer n1 and n2 doesn't easily generalize to a case with multiple features each of different kind with the current input format for forest_minimize(). Once I fix this generalization (I need to go in more detail) I will apply your other feedbacks as well. Thanks...

backtesting/backtesting.py
ttfreeman: Okay, I have renamed the `auto` to `grid` which is the default method. As you mentioned, we don't need to build param_combos if it is not used in skopt (or likely any non-grid method) .

backtesting/backtesting.py
ttfreeman: No it doesn't. I am assuming there could be variable number of features a user might want to optimize the strategy with, so something like **params would be more suitable?

kernc: Ideally, best evaluated **params:
https://github.com/kernc/backtesting.py/blob/191f4753b2a405b20e30add2267afc69a5b06858/backtesting/backtesting.py#L1311-L1312

backtesting/backtesting.py
kernc: 'auto' is somewhat an [industry standard](https://scikit-learn.org/stable/search.html?q=%27auto%27), indicating that the value to be used depends on some aspect of the passed data.

backtesting/backtesting.py
```diff
@@ -1163,8 +1174,11 @@ def run(self, **kwargs) -> pd.Series:
 
     def optimize(self,
                  maximize: Union[str, Callable[[pd.Series], float]] = 'SQN',
+                 method: str = 'grid',
+                 max_tries: Union[int, float] = 100,
                  constraint: Callable[[dict], bool] = None,
                  return_heatmap: bool = False,
+                 return_optimization: bool = False,
```
kernc, line 1181: I'd prefer `return_heatmap=` and `return_optimization=` be non-exclusive.
* [ ] For `method='skopt'`, `return_heatmap=True` should return a multi-indexed pandas Series with known (evaluated) objective values. I'm quite positive this can be reconstructed from [`OptimizeResult`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.OptimizeResult.html).
* [ ] For both `return_heatmap=True, return_optimization=True`, a _3-tuple should be returned_. (Or in case of `method='grid', return_optimization=True` => error.)

API questions like these are open to discussion, but it's my position that will prevail, so I better be convinced.

backtesting/backtesting.py
```diff
@@ -1221,101 +1237,154 @@ def maximize(stats: pd.Series, _key=maximize):
...
+            to model the expensive to evaluate function (Backtest.run()). The
+            model is improved by sequentially evaluating the expensive function
+            at the next best point. Thereby finding the minimum of func with as
+            few evaluations as possible: 
+            (https://scikit-optimize.github.io/stable/modules/generated/skopt.forest_minimize.html).
+
+            The default value for base_estimator is used (ET), and n_calls = max_tries 
+
+            .. TODO::
+                Improve multiprocessing with joblib.parallel .
```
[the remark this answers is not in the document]
ttfreeman: Good point. skopt.forest_minimize() has a number of input parameters that we do not include Backtest.optimize(), I have only refenced the best estimator function that we use ET by default. In a future release maybe more refined input for optimization can be allowed. For the n_jobs, -1 uses the maximum CPU which I think is the best default.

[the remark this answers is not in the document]
kernc: I was contemplating why you went with `forest_minimize()` (and its ET) over `gp_minimize()` or `gbrt_minimize()`. Welcome your thoughts/insights.

> For the n_jobs, -1 uses the maximum CPU which I think is the best default.

I tentatively agree, but spawning multiple processes is [not without overhead](https://stackoverflow.com/a/32649856/1090455). I'm further worried how joblib does memory management of closure structures. If it's not copy-on-write, this will spell disaster for larger data payloads. And with pickling, `n_jobs != 1` might not even work! I know I've went through some hoops to get multiprocessing working well (https://github.com/kernc/backtesting.py/commit/1faa7e863cd11ff7a26fb7f6bd445095978019ee).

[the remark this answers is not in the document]
ttfreeman: Sorry, I probably screwed it up. I guess I wasn't supposed to push the egg-info folder? Didn't expect would add so many files.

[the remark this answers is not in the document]
kernc: Yeah, I use a global gitignore file with patterns for common build artifacts for all languages I touch, so I don't even notice when a project's gitignore is amiss. :sweat_smile:

This one seems to exclude some common dir's contents, so it should exclude the dirs as well. :thinking: Guess not.
https://github.com/kernc/backtesting.py/blob/252cabdf15fd8827560393cc4eb64d48a5a66bd9/.gitignore#L6-L9

kernc:
We also need to document the new feature. Besides the docstring, I'm thinking a usage example (with plots) in a new trailing section (_Stochastic optimization_ or some such) in [_Parameter Heatmap_](https://github.com/kernc/backtesting.py/blob/master/doc/examples/Parameter%20Heatmap.ipynb) notebook, which is already covering roughly the same thing ...

Welcome to further clean this up a bit, or I will take over.

Really excited about getting this in! :heart_eyes:

ttfreeman:
> We also need to document the new feature. Besides the docstring, I'm thinking a usage example (with plots) in a new trailing section (_Stochastic optimization_ or some such) in [_Parameter Heatmap_](https://github.com/kernc/backtesting.py/blob/master/doc/examples/Parameter%20Heatmap.ipynb) notebook, which is already covering roughly the same thing ...
> 
> Welcome to further clean this up a bit, or I will take over.
> 
> Really excited about getting this in! 😍

I am excited too. Will add the documentation. I also need to have it pass Github Action CI so hopefully you can merge. Looking forward to the next issue I can tackle. Please feel free to assign.

backtesting/backtesting.py
```diff
@@ -1224,101 +1244,151 @@ def maximize(stats: pd.Series, _key=maximize):
...
+
+            self.run(**dict(zip(list(kwargs.keys()), res.x)))
+
+            heatmap_df = pd.DataFrame(res.x_iters, columns=list(kwargs.keys()))
+            heatmap = pd.Series(res.func_vals,
+                                name=maximize_key,
+                                index=pd.MultiIndex.from_frame(heatmap_df))
+
+            if return_optimization and return_heatmap:
+                return self._results, res, heatmap
```
arunavo4, line 1376:
```
Expected type 'Union[Series, Tuple[Series, Series]]', got 'Tuple[Any, Any, Series]' instead 
```
Just a small type hinting issue

arunavo4 (commented):
Awesome work! Suggesting just a type hinting fix. The Speed improvement in Optimization is Huge.

kernc:
@arunavo4 Thanks for giving it a try. It should be even more performant now due to just as adequate models with fewer trees.

@ttfreeman I made a few editorial changes, hope you don't mind. Welcome a see through if anything strikes you as odd or misplaced. Excellent work otherwise. We certainly wouldn't have landed this anytime soon without you, and we likely wouldn't end up with anything as awesome as this! This makes me consider maintaining a "what's new" changelog. It's a game changer. Thanks again!

backtesting/backtesting.py
kernc: Finally opted for no `'auto'`. Switching methods is simple enough.

backtesting/backtesting.py
kernc: Applied checking that params pass `constraint=`. Additionally, we're waiting for https://github.com/scikit-optimize/scikit-optimize/pull/971. Welcome a review.

backtesting/backtesting.py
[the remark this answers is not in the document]
kernc: My trials with `n_jobs=-1` were all running slower. `n_jobs=` in scikit-optimize is only used for building the hyperspace landscape estimators, not for running the expensive objective function. :disappointed:

backtesting/backtesting.py
kernc: It's now `Tuple[Series, Series, scipy.optimize.OptimizeResult]`. Optimize result comes after heatmap. :ok_hand:

kernc (approved):

ttfreeman:
> @arunavo4 Thanks for giving it a try. It should be even more performant now due to just as adequate models with fewer trees.
> 
> @ttfreeman I made a few editorial changes, hope you don't mind. Welcome a see through if anything strikes you as odd or misplaced. Excellent work otherwise. We certainly wouldn't have landed this anytime soon without you, and we likely wouldn't end up with anything as awesome as this! This makes me consider maintaining a "what's new" changelog. It's a game changer. Thanks again!

@kernc Thank you. All good. I enjoyed working on this. Great learning experience. Do you want me to take on the task for maintaining the `what's new` section? Or any other issues, please feel free to assign. Cheers...

ttfreeman:
> Awesome work! Suggesting just a type hinting fix. The Speed improvement in Optimization is Huge.

@arunavo4 Thanks for the feedback.

PR #156, opened 2020-10-13
ENH: Add *annualized* return/volatility/Sharpe/Sortino stats

kernc:
Fixes https://github.com/kernc/backtesting.py/issues/150

Where the measurements (slightly) deviate across implementations (e.g. [empyrical](https://github.com/quantopian/empyrical) from Quantopian; it's what pyfolio uses), ours are the more conservative and statistically superior measures.

@eervin123 This is what I came up with. Welcome a quick look over / comments.


eervin123:
This is great.  I like how you handled the 365 vs 252 days, I also agree with sticking to same method for both calculations.  Strange why empyrical would have used a different method.  I look forward to playing with this and testing it out.  

BTW, on a side note, my family is in Slovenia right now.  They tell me it is an amazing place with wonderful people.  I've never been, I wish I could have gone with them.

kernc:
I think they just went for metrics as initially defined (in their respective papers). Makes sense to be in line with other tools. I can expect further sporadic issues inquiring about values deviating from well-known expectations.

Thanks. It's great for hiking this year. General tourism, I guess, not so much. :smile:

backtesting/backtesting.py
```diff
@@ -1339,7 +1339,7 @@ def _compute_stats(self, broker: _Broker, strategy: Strategy) -> pd.Series:
 
         equity = pd.Series(broker._equity).bfill().fillna(broker._cash).values
         dd = 1 - equity / np.maximum.accumulate(equity)
-        dd_dur, dd_peaks = self._compute_drawdown_duration_peaks(pd.Series(dd, index=data.index))
```
eervin123, line 1342: @kernc I was having some issues with Max DD not showing correctly in the stats and in the plots You can see for yourself in the legend MAX DD is 0 days. https://www.realitysharesadvisors.com/fund_dash/fix_this.html  

image


So I saw this change in the diff for recent pr #156 I added this back and now it works.  `index=data.index` instead of `index=index` 

That fixed the calculation, but the red line is misplaced in plots still for some reason so I'm still digging into that.

Here is an example. 
https://www.realitysharesadvisors.com/fund_dash/fix_this_eth.html

kernc: > I added this back and now it works. `index=data.index` instead of `index=index`

I don't understand it, `index` is set to the same value just a couple lines prior:
https://github.com/kernc/backtesting.py/blob/191f4753b2a405b20e30add2267afc69a5b06858/backtesting/backtesting.py#L1343-L1347
and `_compute_drawdown_duration_peaks()` doesn't seem to touch it in any destructive way. :open_mouth: :confused:

> but the red line is misplaced in plots still for some reason so I'm still digging into that.

Can you verify this works on master/released version? If not, I'd suspect `Backtest.plot(resample=True)`:
https://github.com/kernc/backtesting.py/blob/191f4753b2a405b20e30add2267afc69a5b06858/backtesting/backtesting.py#L1479-L1489

eervin123: Okay, so you were absolutely right, the stats were not the problem.  I'm sorry for not catching that before posting.  The drawdown calculations were fine in the _compute_stats.  The legend in the plot disagrees with the stats and I didn't pick up on that. 

**Testing**: I created a new clean environment and pip installed the master/released version of backtesting.py.  When it is run on daily plots or other plots with fewer than 10,000 bars, the equity plot correctly shows the max drawdown in the correct location as well as the red bar across the proper peak/valley/peak.  

However, when it is run on plots that need to be resampled it comes up with some erroneous results.  The 0-day drawdown for example, or the drawdown period over a different section than the max drawdown.  "red dot"

Also of note, though I don't think it matters for the drawdown calculations. I set resample=False in bt.plot(resample=False) and it appears to have ignored the suggestion. 

I'm not sure but is the following if statement ignoring the False boolean? I believe Bools are a subclass of int so checking for str skips down to the "else" statement and reassigns freq to one of the resample periods.  Again, if I'm just causing more trouble than it's worth I'm happy to just blindly point out issues and not even bother trying to identify the culprits. 

`def _maybe_resample_data(resample_rule, df, indicators, equity_data, trades):
    if isinstance(resample_rule, str):
        freq = resample_rule
    else:
        if...`

kernc: > or the drawdown period over a different section than the max drawdown. "red dot"

The max drawdown _period length_ is fully distinct from the max drawdown _value_. But your plots indicate clearly the computations are amiss.

> I'm not sure but is the following if statement ignoring the False boolean? I believe Bools are a subclass of int so checking for str skips down to the "else" statement and reassigns freq to one of the resample periods. Again, if I'm just causing more trouble than it's worth I'm happy to just blindly point out issues and not even bother trying to identify the culprits.

Oops. You're perfectly right. Fixed in https://github.com/kernc/backtesting.py/commit/05329deaa507e64dbd7858f9d253eecb20ab9f16.

I think the one thing everyone likes even better than issues is fixes for them. :stuck_out_tongue_closed_eyes: 

-----

I gather these are unrelated to this PR then? Would you mind opening a new issue for that invalid resampled drawdown?

eervin123:
Sorry for the naive question, I promise I spent a bit of time on google first. Generally speaking about open source projects, when do the changes in a github open source repo appear in the PyPI for pip installs?  

Is there some method for checking based on the version number against a commit number or something?

I assume it is up to the maintainer of the repo on when they want to build the package and send it to PyPI but in this case I just happened to be installing it and noticed that the changes from this commit https://github.com/kernc/backtesting.py/commit/dfadfd7435518ba0f3a08eca2d306113a3cad887

are in the latest pip install, but not the changes from this commit 
https://github.com/kernc/backtesting.py/commit/028f02d4c5c358e39d67411eeb1f1a5f82370cef 

That wouldn't normally jump out to me, but they were both committed on the same day so I just wasn't sure if there is a way to check which version is in PyPI.

kernc:
It's just about [tagging commits](https://github.com/kernc/backtesting.py/tags) in the commit history as released versions. GitHub so far [doesn't show tags in the history](https://github.com/kernc/backtesting.py/commits/master), but other git tools might:
Screenshot_2020-10-31_22-19-31


I arbitrarily decided the changed annualization commit should wait a few days for the remaining three open PRs and then, in accordance with [semver](https://semver.org/), become 0.3.0.

eervin123:
Got it.  Thanks.  I figured it was something like that. I was able to install from the repo so all good there.

Issue #197, opened 2020-12-14
Tracking id from order to trade


qacollective:
### Expected Behavior
It is possible to track which order resulted in which trade.

### Actual Behavior
Once an order is filled, it is discarded.  This is fine, especially to save on resource usage, but there is no way to track the relationships between orders and trades over the long term.

### Steps to Reproduce
1. Run a long backtest which will result in multiple trades being taken out by multiple orders at the same price and tp/sl
2. Attempt to record exactly what happened during trading for later dissection and detailed reporting

### Additional info
Background on this is that all our usual trading software traces an order with an order `id` and that `id` persists throughout all the other record types.  In backtesting.py's case, this is `trades` and `closed_trades`.  If similar tracking was possible in backtesting.py, that would enable:
- a wealth of existing tools to be used with the data that can be output from this solution
- more complex contingent trading algorithms to be tested in backtesting.py
- additional data to be stored about each order or trade, depending on trading method

### Proposed change
I am willing to submit a pull request containing changes to:
- Enable an optional integer parameter for `strategy.buy()`, `strategy.sell()` and `Trade()`, which would be an `integer` (? or any data type? ) named `id`
- This user supplied number would be accessible as another parameter on both `Order` and `Trade`
- This user supplied number would be transferred to a `Trade` when an order is filled and subsequently discarded

However before I do this, I just wanted to check that people (and particularly the author @kernc) think it is a good idea and that there isn't a better way of solving this problem.

### Version info
- Backtesting version: 0.3.0

Devjeel:
I was trying something very similar recently, Where I have to change stoploss from active trades for only one of the trades. 

Whould be best if `buy() / sell()` functions takes optional argument of OrderID, that ultimately goes into `Trades`. With that, later I can modify `sl, tp`, or cancel the open position(ie. if on long, sell instantly).

qacollective:
> I was trying something very similar recently, Where I have to change stoploss from active trades for only one of the trades.

Yes, it is quite difficult to find an order that resulted in a particular trade.  There is no `order_price` on `trade` that could be used to figure that out either.  So an id would solve multiple problems, including tracking extra detail outside of `backtesting.py`.

kernc:
This looks like a duplicate/related to issue https://github.com/kernc/backtesting.py/issues/53.

Can you give an example (code) of some decision/analysis done upon orders/trades with such an id? 

I'm definitely more for the arbitrarily-typed label/tag than a strictly numeric id.

[...]

PR #198, opened 2020-12-15
Add an ID which persists from Order to Trade

qacollective:
I created this PR quickly as a further contribution for discussion on issue #197.  I have not yet had time to test it, but will do first thing tomorrow, as it will make quite a significant improvement to what I'm attempting to do with backtesting.py right now.

I notice that @kernc keeps the code strictly typed, but in this particular instance I made the decision to be type agnostic, enabling the user of this feature to use any data type they want for an id.

I hope people think this is a good idea :worried: :sweat_smile:

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -193,7 +193,8 @@ class Strategy(metaclass=ABCMeta):
             limit: float = None,
             stop: float = None,
             sl: float = None,
-            tp: float = None):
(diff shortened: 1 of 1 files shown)


qacollective:
Going to try to close this pull request and submit a new one, without all the bugs and typeo's now I've had time to actually test it!  Eeep ... this one is an embarassment!  That's what you get for trying to code around kids ...

PR #199, opened 2020-12-16
Update to enable tracking id from order to trade

qacollective:
This is a simple change to backtesting.py to enable a user to track which trade came from which order (and a lot of other potential benefits) as outlined in issue #197.

It simply allows the assignment of an id when creating an order at with `Strategy.buy()` and `Strategy.sell()`.  The id is type agnostic, so the user can make it a string, int, float or whatever may be necessary.

I could not think of a better way to solve this problem and I will be able to test that this works in the next day or so.  I hope people think this is a good idea :worried: :sweat_smile:

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -914,7 +914,12 @@ class _Broker:
 
             # Open a new trade
             if need_size:
-                self._open_trade(adjusted_price, need_size, order.sl, order.tp, time_index, order.id)
+                self._open_trade(adjusted_price,
+                                 need_size,
+                                 order.sl,
+                                 order.tp,
+                                 time_index,
+                                 order.id)
 
                 # We need to reprocess the SL/TP orders newly added to the queue.
                 # This allows e.g. SL hitting in the same bar the order was open.
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -47,6 +47,7 @@ class Strategy(metaclass=ABCMeta):
     `backtesting.backtesting.Strategy.next` to define
     your own strategy.
     """
+
     def __init__(self, broker, data, params):
         self._indicators = []
         self._broker: _Broker = broker
@@ -194,7 +195,7 @@ class Strategy(metaclass=ABCMeta):
             stop: float = None,
             sl: float = None,
             tp: float = None,
-            id = None):
+            tag: object = None):
         """
         Place a new long order. For explanation of parameters, see `Order` and its properties.
 
@@ -202,7 +203,7 @@ class Strategy(metaclass=ABCMeta):
         """
         assert 0 < size < 1 or round(size) == size, \
             "size must be a positive fraction of equity, or a positive whole number of units"
-        return self._broker.new_order(size, limit, stop, sl, tp, id)
+        return self._broker.new_order(size, limit, stop, sl, tp, tag)
 
     def sell(self, *,
              size: float = 1 - sys.float_info.epsilon,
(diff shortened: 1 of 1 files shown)


backtesting/backtesting.py
```diff
@@ -193,30 +193,32 @@ def buy(self, *,
             limit: float = None,
             stop: float = None,
             sl: float = None,
-            tp: float = None):
+            tp: float = None,
+            id = None):
```
kernc, line 197: PEP 8 proposes no spaces around equals if no typing annotation. Annotate with `object` or `typing.Any`.

backtesting/backtesting.py
```diff
@@ -374,7 +376,8 @@ def __init__(self, broker: '_Broker',
                  stop_price: float = None,
                  sl_price: float = None,
                  tp_price: float = None,
-                 parent_trade: 'Trade' = None):
+                 parent_trade: 'Trade' = None,
+                 order_id = None):
```
kernc, line 380: Let's call the parameter `tag=` because tags don't need to be unique.

backtesting/backtesting.py
```diff
@@ -468,6 +473,14 @@ def tp(self) -> Optional[float]:
     def parent_trade(self):
         return self.__parent_trade
 
+    @property
+    def id(self) -> Optional:
+        """
+        An ID attribute which, if set, persists to enable tracking of this order
+        if it becomes a trade in `Strategy.trades` and when closed in `Strategy.closed_trades`.
```
kernc, line 480: Maybe:

> An arbitrary tag object which, if set, propagates to the `Trade` this order started.

Which begs the question, what about orders that _close_ existing trades?

backtesting/backtesting.py
```diff
@@ -511,10 +524,12 @@ def __init__(self, broker: '_Broker', size: int, entry_price: float, entry_bar):
         self.__exit_bar: Optional[int] = None
         self.__sl_order: Optional[Order] = None
         self.__tp_order: Optional[Order] = None
+        self.__order_id = id
 
     def __repr__(self):
         return f'<Trade size={self.__size} time={self.__entry_bar}-{self.__exit_bar or ""} ' \
-               f'price={self.__entry_price}-{self.__exit_price or ""} pl={self.pl:.0f}>'
+               f'price={self.__entry_price}-{self.__exit_price or ""} pl={self.pl:.0f}' \
+               f'{" id="+str(self.__order_id) if self.__order_id is not None else ""}>'
```
kernc, line 532:
```suggestion
               f' id={self.__order_id if self.__order_id is not None else ""}>'
```
... if not moving it out of the return statement / making it optional altogether.

kernc (changes requested):
Will additionally need a unit test and a good example use for the docs.

backtesting/backtesting.py
qacollective: Okay, I'll go with `object` :+1:

backtesting/backtesting.py
qacollective: Great idea, renaming all to `tag` now...

backtesting/backtesting.py
qacollective: But wouldn't that potentially leave a hanging `id=` ?

qacollective:
Sorry, I have completely butchered the PR process here.  git is driving me a bit nuts.

backtesting/backtesting.py
kernc: Ah, that's exactly correct. Never mind then. :flushed:

kernc:
Fyi, you can `git push --force` to overwrite existing branch contents.

qacollective:
> Fyi, you can `git push --force` to overwrite existing branch contents.

:sweat_smile: sorry, yes I'm in the difficult position of trying to do a lot with a little time at the moment, so your advice is great and something I would normally have googled to self-educate but December is :vomiting_face: haha

Issue #204, opened 2020-12-22
Possible bug bt.plot(show_legend=False)


eervin123:
### Expected Behavior
Plot output to browser should look almost the same with or without `bt.plot(show_legend=False)`.

When I run `bt.plot()` my output looks as expected. 
![image](https://user-images.githubusercontent.com/32274861/102834206-3c4a4880-43a8-11eb-9043-38fe93acc658.png)

### Actual Behavior
When run `bt.plot(show_legend=False)` my output is slow to load and comes out distorted.  
example picture here. 
![image](https://user-images.githubusercontent.com/32274861/102834272-71569b00-43a8-11eb-9af2-83b1800b46ae.png)

### Steps to Reproduce
```
from backtesting import Backtest, Strategy
from backtesting.lib import crossover

from backtesting.test import SMA, GOOG


class SmaCross(Strategy):
    n1 = 10
    n2 = 20

    def init(self):
        close = self.data.Close
        self.sma1 = self.I(SMA, close, self.n1)
        self.sma2 = self.I(SMA, close, self.n2)

    def next(self):
        if crossover(self.sma1, self.sma2):
            self.buy()
        elif crossover(self.sma2, self.sma1):
            self.sell()


bt = Backtest(GOOG, SmaCross,
              cash=10000, commission=.002,
              exclusive_orders=True)

output = bt.run()
bt.plot(show_legend=False)
```

### Additional info
I am using backtesting version 0.3.0
bokeh is version 2.2.1
Browser is chrome

kernc:
Apparently [Bokeh Legend location](https://docs.bokeh.org/en/latest/docs/user_guide/styling.html#location) is no longer (:grimacing:) supported to be None the way we are setting it:
https://github.com/kernc/backtesting.py/blob/bbcb7bac8ffe65a5002edfcdf8bc50f92eaca2b9/backtesting/_plotting.py#L616-L618

I think this might now have to be [`.visible`](https://docs.bokeh.org/en/latest/docs/user_guide/styling.html#visible-property). Want to test/make a PR?

eervin123:
Sure thing.

PR #207, opened 2020-12-24
Plot `show_legend = False` fix

eervin123:
Made changes regarding bug #204 based on update to bokeh no longer using `legend.location = None` I also added a test function to display to browser.  (I've never worked with tests so, I'm sure this was done incorrectly @kernc) your test_params test sent everything to temp files which I didn't quite know how to access so I just created a separate one that sent the output to the browser.

Fixes https://github.com/kernc/backtesting.py/issues/204

### DIFF
--- a/backtesting/test/_test.py
+++ b/backtesting/test/_test.py
@@ -642,6 +642,14 @@ class TestPlot(TestCase):
                 with self.subTest(param=p[0]):
                     bt.plot(**dict([p]), filename=f, open_browser=False)
 
+    def test_hide_legend(self):
+        bt = Backtest(GOOG.iloc[:100], SmaCross)
+        bt.run()
+        with _tempfile() as f:
+            bt.plot(filename=f, show_legend=False)
+            # Give browser time to open before tempfile is removed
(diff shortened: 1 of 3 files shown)


kernc:
> `test_params` test sent everything to temp files

That test is just for coverage, to see nothing blows up.

backtesting/test/_test.py
```diff
@@ -642,6 +642,14 @@ def test_params(self):
                 with self.subTest(param=p[0]):
                     bt.plot(**dict([p]), filename=f, open_browser=False)
 
+    def test_hide_legend(self):
+        bt = Backtest(GOOG.iloc[:100], SmaCross)
+        bt.run()
+        with _tempfile() as f:
+            bt.plot(filename=f, show_legend=False)
+            # Give browser time to open before tempfile is removed
+            time.sleep(10)
```
kernc, line 651:
```suggestion
            time.sleep(5)
```
Waiting is bad enough as it is.

kernc:
Lgtm. Many thanks!

Issue #71, opened 2020-05-20
Sharpe Ratio does not take risk-free rate as a parameter


afzalmushtaque:
Backtesting.py Line #954 

https://github.com/kernc/backtesting.py/blob/7010d68ca3d3593457447e68d19efbf51c14ae72/backtesting/backtesting.py#L952

### Expected Behavior
```py
s.loc['Sharpe Ratio'] = (mean_return - risk_free_return) / (returns.std() or np.nan)
```

### Actual Behavior
```py
s.loc['Sharpe Ratio'] = mean_return / (returns.std() or np.nan)
```

### Additional info
[Sharpe Ratio](https://www.investopedia.com/terms/s/sharperatio.asp)

kernc:
With interest rates effectively zero or negative and with short-term bond yields close to zero, assuming the risk-free rate at 0% was an honest, intended, and not unreasonable simplification.

Does your Sharpe change significantly if you amend it so?

afzalmushtaque:
Yes it does. Risk free rate here in India is ~6%.

kernc:
As a quick workaround, might this fix your Sharpe?

```py
def adjusted_sharpe(stats, riskfree_rate=.06):
    mean_returns = stats['Avg. Trade [%]'] / 100
    return stats['Sharpe Ratio'] / mean_returns * (mean_returns - riskfree_rate)
```

You can also use it to optimize: `Backtest.optimize(..., maximize=adjusted_sharpe)`

afzalmushtaque:
Haha thanks for going the extra mile but that's pretty much what I am already doing. Not a deal breaker for me. However most users kind of expect a risk-free-rate parameter for a Sharpe function coming from other quant packages so I guess adding it avoid some confusion and time spent scouring through the source code to ensure it hasn't been hardcoded to something like 2%.

kernc:
It is since v0.3.2 possible to use [`backtesting.lib.compute_stats`](https://kernc.github.io/backtesting.py/doc/backtesting/lib.html#backtesting.lib.compute_stats) and recompute the statistics with a custom risk-free rate:
```py
stats = Backtest(GOOG, MyStrategy).run()
new_stats = compute_stats(stats=stats, data=GOOG, risk_free_rate=.02)
```

PR #281, opened 2021-03-20
ENH: Extract strategy performance method compute_stats

### LINKED ISSUE #194: Separate statistics for long and short trades?
After the backtest run, we see the statistics. It is very good! Is there any simple way to see that statistics separately for long ans short trades?

crazy25000:
Resolves #194 
Resolves https://github.com/kernc/backtesting.py/issues/71

Extracted methods `compute_drawdown_duration_peaks`, `compute_stats`, and class `_Stats` from `backtesting.py` to a new file so that they're available to users.

This will give us the ability to explicitly get performance metrics and have direct access to the data.

TODO:
- [x] Update unit tests
- [x] `flake8` passes
- [x] `mypy` passes
- [ ] ~~Add long/short trade statistics~~
  - ~~Need info about which metrics we want (pips, duration, wins, losses...etc)~~
- [ ] Update documentation

### DIFF
--- a/backtesting/_stats.py
+++ b/backtesting/_stats.py
@@ -1,9 +1,12 @@
-from typing import List
+from typing import List, TYPE_CHECKING, Union
 
 import numpy as np
 import pandas as pd
 
 from ._util import _data_period
+
+if TYPE_CHECKING:
+    from .backtesting import Strategy, Trade
 
 
 def compute_drawdown_duration_peaks(dd: pd.Series):
@@ -26,15 +29,17 @@ def geometric_mean(returns: pd.Series) -> float:
     returns = returns.fillna(0) + 1
     if np.any(returns <= 0):
         return 0
-
     return np.exp(np.log(returns).sum() / (len(returns) or np.nan)) - 1
 
 
 def compute_stats(
-        trades: List[pd.DataFrame],
+        trades: Union[List['Trade'], pd.DataFrame],
         equity: np.ndarray,
         ohlc_data: pd.DataFrame,
-        risk_free_rate: float = 0) -> pd.Series:
+        strategy_instance: 'Strategy',
+        risk_free_rate: float = 0,
+) -> pd.Series:
+    assert -1 < risk_free_rate < 1
 
     index = ohlc_data.index
     dd = 1 - equity / np.maximum.accumulate(equity)
@@ -46,18 +51,23 @@ def compute_stats(
         'DrawdownDuration': dd_dur},
         index=index)
 
-    trades_df = pd.DataFrame({
-        'Size': [t.size for t in trades],
-        'EntryBar': [t.entry_bar for t in trades],
-        'ExitBar': [t.exit_bar for t in trades],
-        'EntryPrice': [t.entry_price for t in trades],
-        'ExitPrice': [t.exit_price for t in trades],
-        'PnL': [t.pl for t in trades],
-        'ReturnPct': [t.pl_pct for t in trades],
-        'EntryTime': [t.entry_time for t in trades],
-        'ExitTime': [t.exit_time for t in trades],
-    })
-    trades_df['Duration'] = trades_df['ExitTime'] - trades_df['EntryTime']
+    if isinstance(trades, pd.DataFrame):
+        trades_df = trades
+    else:
+        # Came straight from Backtest.run()
+        trades_df = pd.DataFrame({
+            'Size': [t.size for t in trades],
+            'EntryBar': [t.entry_bar for t in trades],
+            'ExitBar': [t.exit_bar for t in trades],
+            'EntryPrice': [t.entry_price for t in trades],
+            'ExitPrice': [t.exit_price for t in trades],
+            'PnL': [t.pl for t in trades],
+            'ReturnPct': [t.pl_pct for t in trades],
+            'EntryTime': [t.entry_time for t in trades],
+            'ExitTime': [t.exit_time for t in trades],
+        })
+        trades_df['Duration'] = trades_df['ExitTime'] - trades_df['EntryTime']
+    del trades
 
     pl = trades_df['PnL']
     returns = trades_df['ReturnPct']
@@ -75,8 +85,8 @@ def compute_stats(
     s.loc['Duration'] = s.End - s.Start
 
     have_position = np.repeat(0, len(index))
-    for t in trades:
-        have_position[t.entry_bar:t.exit_bar + 1] = 1
+    for t in trades_df.itertuples(index=False):
+        have_position[t.EntryBar:t.ExitBar + 1] = 1
 
     s.loc['Exposure Time [%]'] = have_position.mean() * 100  # In "n bars" time, not index time
     s.loc['Equity Final [$]'] = equity[-1]
@@ -116,7 +126,7 @@ def compute_stats(
     s.loc['Avg. Drawdown [%]'] = -dd_peaks.mean() * 100
     s.loc['Max. Drawdown Duration'] = _round_timedelta(dd_dur.max())
     s.loc['Avg. Drawdown Duration'] = _round_timedelta(dd_dur.mean())
-    s.loc['# Trades'] = n_trades = len(trades)
+    s.loc['# Trades'] = n_trades = len(trades_df)
     s.loc['Win Rate [%]'] = np.nan if not n_trades else (pl > 0).sum() / n_trades * 100  # noqa: E501
     s.loc['Best Trade [%]'] = returns.max() * 100
     s.loc['Worst Trade [%]'] = returns.min() * 100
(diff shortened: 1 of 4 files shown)


backtesting/_stats.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # Annualized return and risk metrics are computed based on the (mostly correct)
+    # assumption that the returns are compounded. See: https://dx.doi.org/10.2139/ssrn.3054517
+    # Our annualized return matches `empyrical.annual_return(day_returns)` whereas
+    # our risk doesn't; they use the simpler approach below.
+    annualized_return = (1 + gmean_day_return)**annual_trading_days - 1
+    s.loc['Return (Ann.) [%]'] = annualized_return * 100
+    s.loc['Volatility (Ann.) [%]'] = np.sqrt((day_returns.var(ddof=int(bool(day_returns.shape))) + (1 + gmean_day_return)**2)**annual_trading_days - (1 + gmean_day_return)**(2*annual_trading_days)) * 100  # noqa: E501
+    # s.loc['Return (Ann.) [%]'] = gmean_day_return * annual_trading_days * 100
+    # s.loc['Risk (Ann.) [%]'] = day_returns.std(ddof=1) * np.sqrt(annual_trading_days) * 100
```
crazy25000, line 96: Any reason we shouldn't delete these?

backtesting/_stats.py
```diff
@@ -0,0 +1,133 @@
...
+    # If no drawdown since no trade, avoid below for pandas sake and return nan series
+    if not len(df):
+        return (dd.replace(0, np.nan),) * 2
+    df['duration'] = df['iloc'].map(dd.index.__getitem__) - df['prev'].map(dd.index.__getitem__)
+    df['peak_dd'] = df.apply(lambda row: dd.iloc[row['prev']:row['iloc'] + 1].max(), axis=1)
+    df = df.reindex(dd.index)
+    return df['duration'], df['peak_dd']
+
+
+def compute_stats(data, broker, strategy) -> pd.Series:
```
kernc, line 21: In order to expose this function as API (or, rather, an additional wrapper of this function in `backtesting.lib`), we need to revise its inputs since `broker` is not for public to use. I think somewhat like [proposed](https://github.com/kernc/backtesting.py/issues/194#issuecomment-743762084), with `trades=` and `equity=` from [trade data](https://kernc.github.io/backtesting.py/doc/examples/Quick%20Start%20User%20Guide.html#Trade-data).

kernc (changes requested):
> * [ ] Add long/short trade statistics
>   - Need info about which metrics we want (pips, duration, wins, losses...etc)

I'm not in favor of supplying extra new keys in the stats series. It's long as it is.

Instead the idea is, the curious user will be able to use the newly exposed `backtesting.lib.compute_stats()` to compute stats on _any subset_ of `trades` they are interested in:
```py
stats = bt.run()

long_trades = stats._trades.query('Size > 0')
long_stats = compute_stats(long_trades, stats._equity_curve.Equity, data)
```

backtesting/_stats.py
kernc: They are documentation, mentioned in the comment a few lines above.

kernc:
> ```py
> long_stats = compute_stats(long_trades, stats._equity_curve.Equity, data)
> ```

Maybe even better to expose `backtesting.lib.compute_stats()` such that it takes simply
```py
compute_stats(trades: pd.DataFrame,  # trades subset
              stats: pd.Series,      # previous stats
              risk_free_rate: float = 0) -> pd.Series
```
and obtain equity and OHLC data simply from stats passed in.

kernc:
> obtain equity and OHLC data simply from stats passed in

Particularly OHLC data is potentially too heavy (e.g. pickling for multiprocessing) to bake into returned stats. Better pass that separately.

crazy25000:
> > * [ ]  Add long/short trade statistics
> >   
> >   * Need info about which metrics we want (pips, duration, wins, losses...etc)
> 
> I'm not in favor of supplying extra new keys in the stats series. It's long as it is.
> 
> Instead the idea is, the curious user will be able to use the newly exposed `backtesting.lib.compute_stats()` to compute stats on _any subset_ of `trades` they are interested in:
> 
> ```python
> stats = bt.run()
> 
> long_trades = stats._trades.query('Size > 0')
> long_stats = compute_stats(long_trades, stats._equity_curve.Equity, data)
> ```

I agree, it would be better to provide the information and let users calculate metrics they need. I wasn't sure if we had to add the long/short stats based on the original issue conversation.

backtesting/_stats.py
```diff
@@ -0,0 +1,142 @@
...
+        return 0
+
+    return np.exp(np.log(returns).sum() / (len(returns) or np.nan)) - 1
+
+
+def compute_stats(
+        trades: List[pd.DataFrame],
+        equity: np.ndarray,
+        ohlc_data: pd.DataFrame,
+        risk_free_rate: float = 0) -> pd.Series:
```
kernc, line 37: Good. Now let's expose+document this function publicly in `backtesting.lib`. The user will be able to pass `trades: pd.DataFrame` as obtained by this exact method (`stats._trades`), so you have to figure out a way for that.

backtesting/backtesting.py
```diff
@@ -1177,7 +1179,16 @@ def run(self, **kwargs) -> pd.Series:
...
-            self._results = self._compute_stats(broker, strategy)
+            equity = pd.Series(broker._equity).bfill().fillna(broker._cash).values
+
+            self._results = compute_stats(
+                trades=broker.closed_trades,
+                equity=equity,
+                ohlc_data=self._data,
+                risk_free_rate=0.0
+            )
+            self._results.loc['_strategy'] = strategy
```
kernc, line 1190:
```suggestion
            stats = compute_stats(
                trades=broker.closed_trades,
                equity=equity,
                ohlc_data=self._data,
                risk_free_rate=0.0,
            )
            stats.loc['_strategy'] = strategy
            self._results = stats
```

backtesting/backtesting.py
```diff
@@ -1086,7 +1088,7 @@ def __init__(self,
             exclusive_orders=exclusive_orders, index=data.index,
         )
         self._strategy = strategy
-        self._results = None
+        self._results: Union[pd.Series, None] = None
```
kernc, line 1091:
```suggestion
        self._results: Optional[pd.Series] = None
```

kernc (changes requested):

backtesting/_stats.py
```diff
@@ -0,0 +1,142 @@
...
+    s.loc['Return (Ann.) [%]'] = annualized_return * 100
+    s.loc['Volatility (Ann.) [%]'] = np.sqrt((day_returns.var(ddof=int(bool(day_returns.shape))) + (1 + gmean_day_return)**2)**annual_trading_days - (1 + gmean_day_return)**(2*annual_trading_days)) * 100  # noqa: E501
+    # s.loc['Return (Ann.) [%]'] = gmean_day_return * annual_trading_days * 100
+    # s.loc['Risk (Ann.) [%]'] = day_returns.std(ddof=1) * np.sqrt(annual_trading_days) * 100
+
+    # Our Sharpe mismatches `empyrical.sharpe_ratio()` because they use arithmetic mean return
+    # and simple standard deviation
+    s.loc['Sharpe Ratio'] = np.clip((s.loc['Return (Ann.) [%]'] - risk_free_rate) / (s.loc['Volatility (Ann.) [%]'] or np.nan), 0, np.inf)  # noqa: E501
+    # Our Sortino mismatches `empyrical.sortino_ratio()` because they use arithmetic mean return
+    s.loc['Sortino Ratio'] = np.clip(annualized_return / (np.sqrt(np.mean(day_returns.clip(-np.inf, 0)**2)) * np.sqrt(annual_trading_days)), 0, np.inf)  # noqa: E501
```
kernc, line 112: Now that we factor `risk_free_rate` separately, [Sortino ratio](https://en.wikipedia.org/wiki/Sortino_ratio#Definition) also uses something like it (target rate).

backtesting/lib.py
```diff
@@ -77,6 +78,23 @@ def barssince(condition: Sequence[bool], default=np.inf) -> int:
...
+        equity: np.ndarray,
+        ohlc_data: pd.DataFrame,
+        risk_free_rate: float = 0) -> pd.Series:
+    # TODO: Add details
+    """
+    Computes strategy performance metrics.
+
+    >>> broker = stats._strategy._broker
+    >>> equity = pd.Series(broker._equity).bfill().fillna(broker._cash).values
+    >>> perf = compute_stats(trades=broker.closed_trades, equity=equity, ohlc_data=GOOG)
```
crazy25000, line 92: @kernc passing `stats._strategy._broker.closed_trades` makes it compatible as is. If we want to pass `stats._trades`, we'd need to convert the `DataFrame` to a list of `Trade`.

crazy25000:
Oops :stuck_out_tongue:

kernc:
@crazy25000 Many thanks for your work on this! :moon_cake:

PR #282, opened 2021-03-20
ENH: Add tqdm to skopt optimization

### LINKED ISSUE #280: Use tqdm also for skopt optimization
### Expected Behavior

Get some progress indication when

```py
bt.optimize(..., method='skopt')
```

Reported in: https://github.com/kernc/backtesting.py/discussions/267

- Backtesting version: 0.3.1

crazy25000:
Resolves #280 

Added `tqdm` progress bar to `skopt` optimization runs .

TODO:
- [x] Review/test in notebook + terminal (briefly tested in IDE)
  - [x] Jupyterlab `3.0.12`
  - [x] Jupyter notebook `6.2.0`
  - [x] Terminal
- [ ] ~~Write new test to account for failing coverage threshold~~

Screenshot in Jupyterlab. `Skopt optimizations` is for the change in this PR, plays nicely with other progress bars if you use them.

image

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -39,6 +39,18 @@ __pdoc__ = {
 }
 
 
+class SkoptProgressBar:
+    """
+    Progress bar using tqdm for scikit-optimize
+    Open feature request: https://github.com/scikit-optimize/scikit-optimize/issues/674
+    """
+    def __init__(self, **kwargs):
+        self._tqdm = _tqdm(**kwargs)
+
+    def __call__(self, result):
+        self._tqdm.update()
+
+
 class Strategy(metaclass=ABCMeta):
     """
     A trading strategy base class. Extend this class and
@@ -1446,7 +1458,7 @@ class Backtest:
                     kappa=3,
                     n_initial_points=min(max_tries, 20 + 3 * len(kwargs)),
                     initial_point_generator='lhs',  # 'sobel' requires n_initial_points ~ 2**N
-                    callback=DeltaXStopper(9e-7),
+                    callback=[DeltaXStopper(9e-7), SkoptProgressBar(desc="Sequential optimisation using decision trees.")],
                     random_state=random_state)
 
             stats = self.run(**dict(zip(kwargs.keys(), res.x)))
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -1448,7 +1448,7 @@ class Backtest:
                     kappa=3,
                     n_initial_points=min(max_tries, 20 + 3 * len(kwargs)),
                     initial_point_generator='lhs',  # 'sobel' requires n_initial_points ~ 2**N
-                    callback=[DeltaXStopper(9e-7)],
+                    callback=DeltaXStopper(9e-7),
                     random_state=random_state)
 
             stats = self.run(**dict(zip(kwargs.keys(), res.x)))
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -1352,7 +1352,7 @@ class Backtest:
                     with ProcessPoolExecutor() as executor:
                         futures = [executor.submit(Backtest._mp_task, backtest_uuid, i)
                                    for i in range(len(param_batches))]
-                        for future in _tqdm(as_completed(futures), total=len(futures)):
+                        for future in _tqdm(as_completed(futures), total=len(futures), description="Backtest.grid"):
                             batch_index, values = future.result()
                             for value, params in zip(values, param_batches[batch_index]):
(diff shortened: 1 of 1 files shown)


backtesting/backtesting.py
```diff
@@ -39,6 +39,18 @@ def _tqdm(seq, **_):
 }
 
 
+class SkoptProgressBar:
```
crazy25000, line 42: This file is getting pretty big, should we move this to `utils` instead?

backtesting/backtesting.py
```diff
@@ -1446,7 +1458,10 @@ def objective_function(**params):
                     kappa=3,
                     n_initial_points=min(max_tries, 20 + 3 * len(kwargs)),
                     initial_point_generator='lhs',  # 'sobel' requires n_initial_points ~ 2**N
-                    callback=DeltaXStopper(9e-7),
+                    callback=[
+                        DeltaXStopper(9e-7),
+                        SkoptProgressBar(total=max_tries, desc="Skopt optimizations")
```
kernc, line 1463: What if instead of the callback machinery, we made a simple `progress` _variable_ and via closure call its `.update()` in `objective_function()` above (ideally at the start as skopt has its issues)? That way, we don't need a whole class for it.

crazy25000: I think that should work too, going to test it. I used the same implementation as their timer class and was thinking of just switching it out when they implement an official callback. No biggie :p 

https://github.com/scikit-optimize/scikit-optimize/blob/530da127c0e3d92fc5018115585e73fecced12a5/skopt/callbacks.py#L116-L141

backtesting/backtesting.py
```diff
@@ -1420,9 +1420,12 @@ def _optimize_skopt() -> Union[pd.Series,
 
             # np.inf/np.nan breaks sklearn, np.finfo(float).max breaks skopt.plots.plot_objective
             INVALID = 1e300
+            skopt_pbar = _tqdm(dimensions, total=max_tries, desc="Skopt optimizations")
 
             @use_named_args(dimensions=dimensions)
             def objective_function(**params):
+                if hasattr(skopt_pbar, "update"):
```
kernc, line 1427: When does the progress bar not have '.update()`?
https://github.com/tqdm/tqdm/blob/bcce20f771a16cb8e4ac5cc5b2307374a2c0e535/tqdm/std.py#L1195-L1207

I don't think the [CI failures](https://github.com/kernc/backtesting.py/runs/2160506802) were related (:cold_sweat:). Prior commit was the better one. And we can't pass [`dimensions` as the iterable](https://github.com/tqdm/tqdm#documentation); it's `max_tries` that counts.

crazy25000: It fails if `tqdm` is not installed and since it's not listed in the requirements or setup, it won't be installed by default: https://github.com/kernc/backtesting.py/blob/73e1534428fc2db9e681030862133b5067508520/backtesting/backtesting.py#L24-L29 

I can add it as a requirement and get rid of the import/attr checks. I assumed you're trying to keep requirements to a minimal.

crazy25000: > I don't think the CI failures were related (cold_sweat). Prior commit was the better one. And we can't pass dimensions as the iterable; it's max_tries that counts.

Yup, `dimensions` was only used when tqdm isn't installed so that when the default method is executed, it has a value for `seq`. Otherwise it would error out.
```python
def _tqdm(seq, **_): 
         return seq 
```

By using `skopt_pbar.update(1)`, it no longer uses dimensions as the iterable. This way it accounted for both cases. I think it would be better if we explicitly add `tqdm` as a required package.

kernc: Ah, right. I like being lightweight on dependencies. Even as tqdm is lightweight itself, I'd rather something else pulls it, and we just hook into using it.

The we should give our mock tqdm the missing method so as to rely on the uniform, tqdm interface. Simpler still, something like this might work:
```py
progress = iter(_tqdm(repeat(0), total=max_tries))

def objective(...):
    next(progress)
    ...
```

backtesting/backtesting.py
```diff
@@ -1420,9 +1420,12 @@ def _optimize_skopt() -> Union[pd.Series,
 
             # np.inf/np.nan breaks sklearn, np.finfo(float).max breaks skopt.plots.plot_objective
             INVALID = 1e300
+            skopt_pbar = _tqdm(dimensions, total=max_tries, desc="Skopt optimizations")
```
kernc, line 1423: Let's call this description `'Backtest.optimize'` as this is the method the user sees running. And add a similar description for grid tqdm loop above.

crazy25000:
@kernc no worries, we're all busy doing something :smile:

Issue #283, opened 2021-03-20
Resampling error: Multiprocessing AttributeError: '_Indicator' object has no attribute '_opts'


emiliobasualdo:
Hi @kernc, great simulator.
### Expected Behavior
I trying to test a strategy `for` many symbols `[btc-usd, eth-usd, ...]` while multiprocessing.
I believe the simulation is executed successfully. Now, the plotting function seems to be generating an error when Multiprocessing and plotting the Indicator.
When running a simple SMA simulation with `ThreadPoolExecutor`, the program executes & plots perfectly.
When running a simple SMA simulation with `ProcessPoolExecutor`, the plotting produces the following error.

After playing around with the `bt.plot()` parameters, I found out that the error only jumps when the `resample` parameter is set to a value different than True. Example below.

I'm having some hard time to understand why the `_opts` attribute is not assigned in `_Array.__new__` https://github.com/kernc/backtesting.py/blob/5e1acccb7b0c9c1f369531035a9c887736f92cc5/backtesting/_util.py#L48-L52

### Actual Behavior
```
Traceback (most recent call last):
  File "/Users/pilo/development/itba/pf/test/sma.py", line 94, in <module>
    bt.plot(plot_equity=True, superimpose=False, resample="5T",
  File "/Users/pilo/development/itba/pf/test/venv/lib/python3.8/site-packages/backtesting/backtesting.py", line 1704, in plot
    plot(
  File "/Users/pilo/development/itba/pf/test/venv/lib/python3.8/site-packages/backtesting/_plotting.py", line 185, in plot
    df, indicators, equity_data, trades = _maybe_resample_data(
  File "/Users/pilo/development/itba/pf/test/venv/lib/python3.8/site-packages/backtesting/_plotting.py", line 110, in _maybe_resample_data
    indicators = [_Indicator(i.df.resample(freq, label='right').mean()
  File "/Users/pilo/development/itba/pf/test/venv/lib/python3.8/site-packages/backtesting/_plotting.py", line 110, in <listcomp>
    indicators = [_Indicator(i.df.resample(freq, label='right').mean()
  File "/Users/pilo/development/itba/pf/test/venv/lib/python3.8/site-packages/backtesting/_util.py", line 83, in df
    df = pd.DataFrame(values.T, index=self._opts['data'].index,
AttributeError: '_Indicator' object has no attribute '_opts'

``` 

### Steps to Reproduce
I have made a very simplified version by forking the GOOG data with SMA twice.
```py
[...]

PR #285, opened 2021-03-21
BUG: Handle _Array properties when pickling

kernc:
Fixes https://github.com/kernc/backtesting.py/issues/283 by (un-)pickling `_Array` properties also.

~~The remaining issue is `_Array._opts['data']: pd.DataFrame`, which is the original OHLC data, which might prove a strain to repeatedly (de-)serialize.~~ Worked around by saving only `.index` for 1/5 of the weight.


crazy25000 (approved):

kernc:
Thanks @crazy25000 for reviewing.

PR #300, opened 2021-04-04
Commission constraints for rebates

### LINKED ISSUE #248: Would it be possible to add a positive commission?
I'm trying to backtest a limit orders based strategy on a crypto exchange named bybit. The problem is that you get rebates when you use limit orders. But I see I can't enter positive commissions (-0.00025). Would it be possible to change this?

```
AssertionError: commission should be between 0-10%, is -0.00025
```

muhrifqii:
Fixes #248

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -664,8 +664,9 @@ class _Broker:
     def __init__(self, *, data, cash, commission, margin,
                  trade_on_close, hedging, exclusive_orders, index):
         assert 0 < cash, f"cash should be >0, is {cash}"
-        assert -.1 <= commission < .1, (f"commission should be between -10% "
-            "(e.g. market-maker's rebates) and 10% (fees), is {commission}")
+        assert -.1 <= commission < .1, \
+            ("commission should be between -10% "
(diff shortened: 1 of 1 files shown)


backtesting/backtesting.py
```diff
@@ -664,7 +664,8 @@ class _Broker:
     def __init__(self, *, data, cash, commission, margin,
                  trade_on_close, hedging, exclusive_orders, index):
         assert 0 < cash, f"cash should be >0, is {cash}"
-        assert 0 <= commission < .1, f"commission should be between 0-10%, is {commission}"
+        assert -.1 <= commission < .1, (f"commission should be between -10% "
+            "(e.g. market-maker's rebates) and 10% (fees), is {commission}")
```
kernc, line 668: _This_ string literal should be the f-string. It's what's tripping the CI.

muhrifqii: yeah I put on the wrong position, sorry for polluting commit history. It should be squashed 😄

kernc:
Thanks for taking the time to patch this. Hope it serves you well.

PR #319, opened 2021-04-27
AssertionError while using TrailingStrategy kernc#316 

zlpatel:
I have found the culprit. It's `self.__atr[-1]`.

> /usr/local/lib/python3.8/site-packages/backtesting/lib.py(433)set_atr_periods()
> -> self.__atr = atr
> (Pdb) pp atr[-1]
> 62.03802490234375
> 

> /usr/local/lib/python3.8/site-packages/backtesting/lib.py(440)set_trailing_sl()
> -> self.__n_atr = n_atr
> (Pdb) pp n_atr
> 1.5

> /usr/local/lib/python3.8/site-packages/backtesting/lib.py(446)next()
> -> trade.sl = max(trade.sl or -np.inf,
> (Pdb) pp self.data.Close[-1]
> 22.170000076293945

so per the formula: `trade.sl = max(trade.sl or -np.inf, 
                self.data.Close[-1] - self.__atr[-1] * self.__n_atr)`

22.17 - (62.04*1.5) = -70.89

the code shouldn't be using atr[-1]. It should be using atr[index] where index = len(self.data)-1

Because, when `self.data` in `init()` is different from self.data in `next()`. `self.data` in `init()` has all the data. while the one in `next()` gets accumulated data over iterations. i.e. 5 iterations -> 5 data. n iteration -> n data. so doing `self.data.Close[-1]` in `next()` gets you the last record. but the same is not true for `self.__atr`, because `self.__atr` always has full computed set from `init()`, so you need to use absolute index while accessing atr value from `self.__atr`.


kernc:
Continued in https://github.com/kernc/backtesting.py/pull/322.

Issue #316, opened 2021-04-26
AssertionError while using TrailingStrategy


zlpatel:
I am seeing this error when try to run backtest with 10000 days worth data. It doesn't happen for all the stocks, but some. e.g AAPL, AMZN.

```
Index: 1946  Close: 35.88999938964844  ATR: 1.6433353276970901
Index: 1947  Close: 35.779998779296875  ATR: 1.5673829350797757
Index: 1948  Close: 35.779998779296875  ATR: 1.5232842084983242
Index: 1949  Close: 36.029998779296875  ATR: 1.4951925556138177
Index: 1950  Close: 36.13999938964844  ATR: 1.438393141851363
Index: 1951  Close: 35.65999984741211  ATR: 1.3906505347952947
Traceback (most recent call last):
  File "backtesting_usa.py", line 116, in <module>
    output = bt.run()
  File "/usr/local/lib/python3.8/site-packages/backtesting/backtesting.py", line 1165, in run
    strategy.next()
  File "backtesting_usa.py", line 83, in next
    super().next()
  File "/usr/local/lib/python3.8/site-packages/backtesting/lib.py", line 446, in next
    trade.sl = max(trade.sl or -np.inf,
  File "/usr/local/lib/python3.8/site-packages/backtesting/backtesting.py", line 633, in sl
    self.__set_contingent('sl', price)
  File "/usr/local/lib/python3.8/site-packages/backtesting/backtesting.py", line 652, in __set_contingent
    assert price is None or 0 < price < np.inf
AssertionError
```

I tried printing the index values for each iteration and I found that for this dataset, it processed till index 1951 and failed on 1952.

If I reduce the backtesting period to 5000, it works because it doesn't come across that piece of data (index 1952). I checked that data and verified the close price, and calculated ATR(14). 

Below is the code snippet:

```py
class Breakout(TrailingStrategy):
	timeperiod = 10
	position_size_decimal = 0.2
	min_close = []
	max_close = []
	def init(self):
		super().init()
		self.ema20 = self.I(EMA,self.data.Close,20,overlay=True)
		self.atr14 = self.I(ATR,self.data.High,self.data.Low,self.data.Close,14)
		self.set_atr_periods(14)
		self.set_trailing_sl(1)
		print(type(self.data.Close))
		self.min_close, self.max_close = MINMAX(self.data.Close, timeperiod=self.timeperiod)

	def next(self):
		super().next()
		index = len(self.data)-1
[...]

PR #322, opened 2021-04-27
 AssertionError while using TrailingStrategy

zlpatel:
Fixes https://github.com/kernc/backtesting.py/issues/316

I have found the culprit. It's `self.__atr[-1]`.

> /usr/local/lib/python3.8/site-packages/backtesting/lib.py(433)set_atr_periods()
> -> self.__atr = atr
> (Pdb) pp atr[-1]
> 62.03802490234375
> 

> /usr/local/lib/python3.8/site-packages/backtesting/lib.py(440)set_trailing_sl()
> -> self.__n_atr = n_atr
> (Pdb) pp n_atr
> 1.5

> /usr/local/lib/python3.8/site-packages/backtesting/lib.py(446)next()
> -> trade.sl = max(trade.sl or -np.inf,
> (Pdb) pp self.data.Close[-1]
> 22.170000076293945

so per the formula: `trade.sl = max(trade.sl or -np.inf, 
                self.data.Close[-1] - self.__atr[-1] * self.__n_atr)`

22.17 - (62.04*1.5) = -70.89

the code shouldn't be using atr[-1]. It should be using atr[index] where index = len(self.data)-1

Because, when `self.data` in `init()` is different from self.data in `next()`. `self.data` in `init()` has all the data. while the one in `next()` gets accumulated data over iterations. i.e. 5 iterations -> 5 data. n iteration -> n data. so doing `self.data.Close[-1]` in `next()` gets you the last record. but the same is not true for `self.__atr`, because `self.__atr` always has full computed set from `init()`, so you need to use absolute index while accessing atr value from `self.__atr`.


backtesting/test/__init__.py
```diff
@@ -15,6 +15,8 @@ def _read_file(filename):
 EURUSD = _read_file('EURUSD.csv')
 """DataFrame of hourly EUR/USD forex data from April 2017 to February 2018."""
 
+AMZN = _read_file('AMZN.csv')
+"""DataFrame of daily NASDAQ:AMZN (Amazon) stock price data from 15/05/1997 to 23/04/2021."""
```
kernc, line 19: I'm quite opposed to vendoring a new data set in a bug fix like this. I meant if you're able to easily construct a small, contrived unit test / test case, i.e., something that fails on `atr[-1]`, but not with `atr[index]`. Otherwise, no test is preferred to this. :grimacing:

zlpatel: Yea. It's little difficult to test atr[-1] unless I write that piece of code out in the test case to do the computation But then it won't really be considered as a unit test for the TrailingStrategy. Rather it would be a logic test. Is it okay to skip the test for this one then? I can add descriptive comment in that line of code as to why I am using index instead of -1.

backtesting/lib.py
```diff
@@ -441,13 +441,21 @@ def set_trailing_sl(self, n_atr: float = 6):
 
     def next(self):
         super().next()
+        """
+        self.data in init() is different from self.data in next(). self.data in init() has all the data. 
+        while the one in next() gets accumulated data over iterations. i.e. 5 iterations -> 5 data. n iteration -> n data. 
+        so doing self.data.Close[-1] in next() gets you the last record. 
+        but the same is not true for self.__atr, because self.__atr always has fully computed set from init(), 
+        so we need to use absolute index while accessing atr value from self.__atr.
+        """
```
kernc, line 450: Infer _truth_ from the code; only comment-document the exceptions:
```suggestion
        # Can't use index=-1 because self.__atr is not an Indicator type
```

zlpatel:
@kernc It passed all the checks! What are the next steps?

kernc:
Merging. Many thanks for taking the time to investigate and fix this! :cake:

fabge:
I thought this PR was merged, was there a rollback?
I am experiencing similiar issues and the above seems to be the fix.

Issue #223, opened 2021-01-08
Trailing pct instead of ATR


riodda:
### Expected Behavior
Allow trailing with PRC instead of ATR

### Actual Behavior

Currently the backtesting allow trailing sels only based on ATR, most of crypto bots works in pct trailing (pct for trailing activation and then pct trailing for position close).

kernc:
This is the whole of [`TrailingStrategy`](https://kernc.github.io/backtesting.py/doc/backtesting/lib.html#backtesting.lib.TrailingStrategy):
https://github.com/kernc/backtesting.py/blob/1ee5670d03b26680d9add5c96a9f98e44fc722a2/backtesting/lib.py#L406-L450
It should be fairly simple to adapt/duplicate one for percent values.

riodda:
Yup, the trailing strategy only supports ATR multiples, but most of the cyrpto bots trail in percentage would be nice to implement as option for trailing strategy so you can choose between ATR or pct for backtesting the best trailing.

kernc:
As a quick workaround, something like this might work:
```py
def pct_to_atr(pct: float, data: pd.DataFrame):
    """Express `pct` change in units of ATR."""
    assert 0 < pct < 1
    h, l, c_prev = data.High, data.Low, pd.Series(data.Close).shift(1) 
    tr = np.max([h - l, (c_prev - h).abs(), (c_prev - l).abs()], axis=0) 
    atr = np.nanmean(tr)
    return (pct * data.Close.iloc[-1]) / atr
```

riodda:
Actually i would need the oppsite, as with backtesting i can find the optimal atr value for trailing the will have to calculate the equivalent pct for program the bot.
I'll find a solution, thanks anyway.

kernc:
Should be rather easy to invert, something like:
```py
def atr_to_pct(n_atr: float, data: pd.DataFrame):
    """Express `n_atr` in units of ATR as percentage of current price."""
    h, l, c_prev = data.High, data.Low, pd.Series(data.Close).shift(1) 
    tr = np.max([h - l, (c_prev - h).abs(), (c_prev - l).abs()], axis=0) 
    atr = np.nanmean(tr)
    return n_atr * atr / data.Close.iloc[-1]
```

zlpatel:
I am working on this enhancement.

zlpatel:
@kernc can you review this change?

PR #377, opened 2021-06-13
 Trailing pct instead of ATR #223 

zlpatel:


Benouare:
Same as the other request. 
On modification/feature/task = one commit.

backtesting/lib.py
```diff
@@ -452,6 +452,43 @@ def next(self):
...
+    def init(self):
+        super().init()
+
+    def set_trailing_sl(self, percentage: float = 5):
+        assert percentage > 0, "percentage must be greater than 0"
+        """
+        Sets the future trailing stop-loss as some (`percentage`)
+        percentage away from the current price.
+        """
+        self._sl_percent = percentage
```
Benouare, line 478: `position.pl_pct ` is in 0.XX. 
To avoid confusion, you may change _sl_percent to _sl_pct, and change the value to percentage/100 ?

backtesting/lib.py
```diff
@@ -452,6 +452,43 @@ def next(self):
...
+    def next(self):
+        super().next()
+        index = len(self.data)-1
+        for trade in self.trades:
+            if trade.is_long:
+                trade.sl = max(trade.sl or -np.inf,
+                               self.data.Close[index]*(1-(self._sl_percent/100)))
+            else:
+                trade.sl = min(trade.sl or np.inf,
+                               self.data.Close[index]*(1+(self._sl_percent/100)))
```
Benouare, line 489: change _sl_percent to _sl_pct and remove /100?

Benouare (changes requested):
Looks good to me. 
I think that you should rebase your commit to have a clean historic, and change the propertie names.

backtesting/lib.py
```diff
@@ -452,6 +452,43 @@ def next(self):
...
+    Remember to call `super().init()` and `super().next()` in your
+    overridden methods.
+    """
+    _sl_pct = 5.
+
+    def init(self):
+        super().init()
+
+    def set_trailing_sl(self, percentage: float = 5):
+        assert percentage > 0, "percentage must be greater than 0"
```
kernc, line 473: Everywhere else, the framework uses "pct" and "percent" to really mean [rate](https://en.wikipedia.org/wiki/Rate_(mathematics)) (e.g. 5% == 0.05). I wonder if here shouldn't be the same.

backtesting/lib.py
```diff
@@ -452,6 +452,43 @@ def next(self):
...
+    def next(self):
+        super().next()
+        index = len(self.data)-1
+        for trade in self.trades:
+            if trade.is_long:
+                trade.sl = max(trade.sl or -np.inf,
+                               self.data.Close[index]*(1-self._sl_pct))
+            else:
+                trade.sl = min(trade.sl or np.inf,
+                               self.data.Close[index]*(1+self._sl_pct))
```
kernc, line 489:
```suggestion
        price = self.data.Close[-1]
        for trade in self.trades:
            if trade.is_long:
                trade.sl = max(trade.sl or -np.inf, price*(1 - self._sl_pct))
            else:
                trade.sl = min(trade.sl or np.inf, price*(1 + self._sl_pct))
```

backtesting/lib.py
zlpatel: So are you saying that a user should be inputting 0.05 instead of 5 and the code should use that value as is?

backtesting/lib.py
zlpatel: Before I make this change, I want to let you know that I used `index = len(self.data)-1` instead of -1 to be consistent with the TrailingStrategy code. I understand that in TrailingStrategy we were using ATR which was computed differently and to get an absolute value for an index, we had to use len(self.data)-1, But I don't see any harm in using it here. Thoughts?

zlpatel:
@kernc Have you had a chance to look at the questions that I asked?

PR #384, opened 2021-06-16
Add Histogram indicator plot style kernc#195

### LINKED ISSUE #195: MACD indicator plotting style
I am using the method `I` to plot MACD, but get the wrong style with histogram.
Maybe need more style settings with the method.

zlpatel:
added an option `histogram=True` for `I()` function. it plots histogram using `vbar` function of bokeh js.

Fixes https://github.com/kernc/backtesting.py/issues/195

### DIFF
--- a/backtesting/_plotting.py
+++ b/backtesting/_plotting.py
@@ -513,6 +513,7 @@ return this.labels[index] || "";
 
             is_overlay = value._opts['overlay']
             is_scatter = value._opts['scatter']
+            is_histogram = value._opts['histogram']
             if is_overlay:
                 fig = fig_ohlc
             else:
@@ -532,7 +533,10 @@ return this.labels[index] || "";
                 tooltips.append(f'@{{{source_name}}}{{0,0.0[0000]}}')
                 if is_overlay:
                     ohlc_extreme_values[source_name] = arr
-                    if is_scatter:
+                    if is_histogram:
+                        fig.vbar('index', BAR_WIDTH, source_name, source=source,
+                                 legend_label=legend_label, color=color)
+                    elif is_scatter:
                         fig.scatter(
                             'index', source_name, source=source,
                             legend_label=legend_label, color=color,
@@ -544,7 +548,10 @@ return this.labels[index] || "";
                             legend_label=legend_label, line_color=color,
                             line_width=1.3)
                 else:
-                    if is_scatter:
+                    if is_histogram:
+                        r = fig.vbar('index', BAR_WIDTH, source_name, source=source,
+                                     legend_label=LegendStr(legend_label), color=color)
+                    elif is_scatter:
                         r = fig.scatter(
                             'index', source_name, source=source,
                             legend_label=LegendStr(legend_label), color=color,
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -75,7 +75,7 @@ class Strategy(metaclass=ABCMeta):
 
     def I(self,  # noqa: E741, E743
           func: Callable, *args,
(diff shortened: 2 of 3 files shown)


Benouare (approved):

AGG2017:
Great idea to add histograms to the charts. I tested that with many combinations of indicators and it perfectly works. Couldn't find an easy way to use factor_cmap to make one indicator with two colors (let say green for + values and red for - values). To add color horizontal lines at fixed levels is much easier. Thanks for the proposal and will wait to see it in the next official updates.

zlpatel:
@AGG2017 thanks for your feedback.

AGG2017:
@zlpatel Thanks for the idea. It is not practical when using standard TA libraries but for some custom indicators is perfect. I had some success using linear_cmap with providing low/high values and their colors. It works good for two colors but is not practical for the squeeze momentum indicator where I will need 4 colors. Anyway, I will try to rewrite all plot parameters for the indicators to have only one dict parameter style, one for each element of the indicator in a list. Then I can select the style for each part of the indicator like for example MACD needs two lines and one histogram.

zlpatel:
> @zlpatel Thanks for the idea. It is not practical when using standard TA libraries but for some custom indicators is perfect. I had some success using linear_cmap with providing low/high values and their colors. It works good for two colors but is not practical for the squeeze momentum indicator where I will need 4 colors. Anyway, I will try to rewrite all plot parameters for the indicators to have only one dict parameter style, one for each element of the indicator in a list. Then I can select the style for each part of the indicator like for example MACD needs two lines and one histogram.

@AGG2017 I see what you're saying. I did think about that but couldn't find an easy way to allow multiple style plots into a single indicator. May be we can explore a little more. May be we need to have a new Indicator class which can allow plotting multiple I() in a single plot.

AGG2017:
@zlpatel I want just to add one more parameter `style` to replace everything else when defined in order to be compatible with all the rest. Let say you have predefined style_macd={'names':['macd','sig','hist'], 'styles':['line','line','histogram'], 'colors':['blue','black','green(+):red(-)']} and then just use in indicator parameters only `style=style_macd`. Something like that. Each element of one complex indicator will have its own style parameter, predefined in advance for the most used one.

vladimircape:
3 month passed but code is still not in official repo :(

quanengineering:
Hi @kernc,

It would be great if you could take a look at this one. I think this is the missing piece that everyone is looking for.

ivaigult:
I appreciate that @kernc does not want to extend the API to allow charts customization (see [this discussion](https://github.com/kernc/backtesting.py/discussions/474)).

> and backtesting.py could never strive to replace TradingView

I tend to agree that it would be hard to extend the existing `I` interface to add all possible charting styles. This can be a nightmare from the maintenance and API perspective. However, this is what the customers want (including myself). So my question is:

Can we tweak the `I` interface somehow s.t. the users get the raw access to bokeh? For instance they can set some "unspecified" or "unchecked" arguments of `I`, and they are just get piped to bokeh as they are. Something like:
```
self.I(my_indicator, self.close, bokeh_fn="vbar", bokeh_args={"width": 42})
```

or something even more flexible, so plotting is completely decoupled from the indicators.

Thoughts?

JamesonDiamond1 (approved):
@JamesonDiamond1

RahmanDanial (changes requested):
Your Color & Scatter has error.

0xWelsh:
If anyone wants a simpler alternative for basic MACD/RSI strategies, I've been using this Gekko fork: https://github.com/0xWelsh/gekko/releases/latest — works out of the box, no Docker needed.

Issue #382, opened 2021-06-15
Need a way to specify meaningful legend names when multiple series values are used in a single Indicator


zlpatel:
### Expected Behavior

There should be a way to specify meaningful legend names when multiple series values are used in a single Indicator

### Actual Behavior

Currently, you can only specify single name for Indicator. But there are cases where you need to plot multiple data series in a single Indicator and it's important to label them for easy identification.

Currently, There's an option to specify list of colors (one for each data series). I think if we allow passing list of legend names (as a separate parameter) that would be great.

### Additional info
<img width="609" alt="image" src="https://user-images.githubusercontent.com/3293739/122130527-9d4d5900-cdec-11eb-8be6-b244f204d6fd.png">

zlpatel:
After my code fix it looks like this: 
<img width="763" alt="image" src="https://user-images.githubusercontent.com/3293739/122137005-8a408600-cdf8-11eb-9952-93a00c6b362e.png">

zlpatel:
@kernc could you mark this as an enhancement?

PR #385, opened 2021-06-16
Allow setting legend names when multiple series values are used in an Indicator #382

zlpatel:
Fixes https://github.com/kernc/backtesting.py/issues/382

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -76,7 +76,7 @@ class Strategy(metaclass=ABCMeta):
     def I(self,  # noqa: E741, E743
           func: Callable, *args,
           name=None, plot=True, overlay=None, color=None, scatter=False,
-          **kwargs) -> np.ndarray:
+          legends=None, **kwargs) -> np.ndarray:
         """
         Declare indicator. An indicator is just an array of values,
         but one that is revealed gradually in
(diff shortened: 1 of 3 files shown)


Benouare (approved):

backtesting/backtesting.py
```diff
@@ -105,6 +105,10 @@ def I(self,  # noqa: E741, E743
         If `scatter` is `True`, the plotted indicator marker will be a
         circle instead of a connected line segment (default).
 
+        `legends` can be list or array of string values to represent
+        legends on your indicator chart. By default it's set to None,
+        and `name` is used as legends.
```
kernc, line 110: Why don't we simply reuse `name=` to accept a list of name values?

zlpatel: I didn't use it thinking it may be used in other places. so I wanted to keep it backward compatible. But if it's not being used elsewhere, I can change my code to use name instead of introducing a new attribute. Let me know.

zlpatel: @kernc did you get a chance to look at this?

aradng (commented):
can we get a merge this been sitting here for 3 months

backtesting/backtesting.py
ivaigult: I am pretty sure, it should be possible to reuse `name=`, since `color=` is used in the same fashion. For instance here is an [example](https://github.com/kernc/backtesting.py/discussions/419#discussioncomment-1033981) how to set multiple colors.

Having two arguments to set one thing is very confusing from the API perspective.

zlpatel: @ivaigult Can you try and let me know. My first attempt was to pass array to name but it didn't work. I already mentioned my reasoning in previous comment. If you can make it work with name that would be great :)

ivaigult: @zlpatel I uploaded my attempt in #980 :slightly_smiling_face:

PR #389, opened 2021-06-17
Auto-scale Y-axis for indicators when zooming #356

### LINKED ISSUE #356: Allow Auto-scaling of Y axis values for Indicator(when overlay=False) when you zoom in or out the chart.
I love backtesting.py and I use it extensively. I love that it can seamlessly integrate pandas-ta, ta-lib etc. and shows chart in html.

But, There's something that I noticed which makes it difficult to view indicator graph on the chart if you're backtesting for longer period. Below is an example of an indicator view on the chart without zooming in. Notice the y axis scales. 
Screenshot 2021-05-21 at 4 03 16 PM

Now, when I zoom in to focus on a specific period when ttm squeeze value stayed between -5 to +5, it still shows the full scale from -15 to+20. which makes it really hard to visualize the indicator chart. See below.
Screenshot 2021-05-21 at 4 09 25 PM

I know that re-scaling is allowed for the price chart/Volume. Is that something that can be implemented on the indicator charts as well?

zlpatel:
Fixes https://github.com/kernc/backtesting.py/issues/356

### DIFF
--- a/backtesting/_plotting.py
+++ b/backtesting/_plotting.py
@@ -502,6 +502,7 @@ return this.labels[index] || "";
 
         ohlc_colors = colorgen()
         indicator_figs = []
+        non_overlay_indicator_idxs = []
 
         for i, value in enumerate(indicators):
             value = np.atleast_2d(value)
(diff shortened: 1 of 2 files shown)


Benouare (approved):

backtesting/_plotting.py
```diff
@@ -607,9 +615,16 @@ def __eq__(self, other):
 
     custom_js_args = dict(ohlc_range=fig_ohlc.y_range,
                           source=source)
+    if plot_pl:
+        custom_js_args.update(pl_range=fig_pl.y_range)
```
zlpatel, line 619: This change was not required. I think I might have added that while I was trying to see if I can add auto scale feature for P/L chart. Since it's already been added in the commit, I still see a value in keeping it for future use, But if you think that it's not worth adding it right now and will cause confusion then I can remove it. Let me know.

zlpatel:
@kernc can you please take a look at this pull request?

JamesonDiamond1 (approved):
@JamesonDiamond1

PR #390, opened 2021-06-18
implemented distance based trailing stop loss

DavyJo:
Implemented fixed price (distance) sl


kernc:
Thanks for the PR. I happen to prefer https://github.com/kernc/backtesting.py/pull/377. With two classes, one can set ATR-based SL and %-based SL simultaneously.

PR #415, opened 2021-07-17
Return plot when plotting output 

sandboiii:
Hello, I had a [discussion](https://github.com/kernc/backtesting.py/discussions/405) about being able to save plots and that it would require to return plot figure to use export_png function. So, I made a pr for this.


kernc:
Thanks! :cake:

PR #483, opened 2021-09-28
simplified multiprocessing on posix systems

stnatter:
added `mp.set_start_method('fork')` for posix systems after imports. Also corrected some typos in doc strings

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -21,6 +21,9 @@ from typing import Callable, Dict, List, Optional, Sequence, Tuple, Type, Union
 import numpy as np
 import pandas as pd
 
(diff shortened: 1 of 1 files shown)


kernc:
I'd prefer to not set/change multiprocessing policy at import time. This is on the user to do, and our warning is in order. (The warning might be expanded, though, with a code snippet and/or a reference link.)

Additionally, the code as suggested now wouldn't work (_at all_) on non-POSIX systems like Windos.

stnatter:
Fair enough! Thanks for the consideration and the fantastic work. Am currently using https://github.com/abbass2/pyqstrat for backtesting strategies and considering to use `backtesting.py` going forward.

PR #579, opened 2022-01-25
added get_closed_trades

ayushd49:


kernc:
Indeed, closed_trades can be obrained via [`Backtest.run()`](https://kernc.github.io/backtesting.py/doc/backtesting/backtesting.html#backtesting.backtesting.Backtest.run)-returned `stats['_trades']`.

PR #635, opened 2022-04-20
Reading data from yahoo

wallacetju:


wallacetju:
Sorry, I was about to merge to the forked one

PR #640, opened 2022-05-08
Remove flooring on ratios and adding Kelly Criterion

t2o2:
* Remove flooring on Sharpe Ratio
* Adding Kelly Criterion

### DIFF
--- a/backtesting/_stats.py
+++ b/backtesting/_stats.py
@@ -117,11 +117,11 @@ def compute_stats(
 
     # Our Sharpe mismatches `empyrical.sharpe_ratio()` because they use arithmetic mean return
     # and simple standard deviation
-    s.loc['Sharpe Ratio'] = np.clip((s.loc['Return (Ann.) [%]'] - risk_free_rate) / (s.loc['Volatility (Ann.) [%]'] or np.nan), 0, np.inf)  # noqa: E501
+    s.loc['Sharpe Ratio'] = (s.loc['Return (Ann.) [%]'] - risk_free_rate) / (s.loc['Volatility (Ann.) [%]'] or np.nan)  # noqa: E501
     # Our Sortino mismatches `empyrical.sortino_ratio()` because they use arithmetic mean return
-    s.loc['Sortino Ratio'] = np.clip((annualized_return - risk_free_rate) / (np.sqrt(np.mean(day_returns.clip(-np.inf, 0)**2)) * np.sqrt(annual_trading_days)), 0, np.inf)  # noqa: E501
+    s.loc['Sortino Ratio'] = (annualized_return - risk_free_rate) / (np.sqrt(np.mean(day_returns.clip(-np.inf, 0)**2)) * np.sqrt(annual_trading_days))  # noqa: E501
     max_dd = -np.nan_to_num(dd.max())
-    s.loc['Calmar Ratio'] = np.clip(annualized_return / (-max_dd or np.nan), 0, np.inf)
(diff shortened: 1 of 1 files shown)


backtesting/backtesting.py
```diff
@@ -1453,17 +1454,29 @@ def objective_function(**params):
...
-                    dimensions=dimensions,
-                    n_calls=max_tries,
-                    base_estimator=ExtraTreesRegressor(n_estimators=20, min_samples_leaf=2),
-                    acq_func='LCB',
-                    kappa=3,
-                    n_initial_points=min(max_tries, 20 + 3 * len(kwargs)),
-                    initial_point_generator='lhs',  # 'sobel' requires n_initial_points ~ 2**N
-                    callback=DeltaXStopper(9e-7),
-                    random_state=random_state)
+                if method == 'skopt_gp':
```
kernc, line 1457: I think these other changes probably don't belong here?

t2o2: Sorry accidental checkin, removed using forced push on my branch

cscbai:
Any other comments on the PR?

t2o2:
Any other comments?

cscbai:
Any update on this review?

kernc:
I've actually been meaning to but just didn't come round to trying it out and seeing what happens when positions are indeed sized as Kelly Criterion recommends. Say I run a backtest and get `Kelly Criterion=0.5`. Say I now repeat the same strategy backtest with trades `size=0.5`. What does the Kelly Criterion say at the end of this backtest? And if the value is different from before, of what real use is it?

t2o2:
The traditional meaning of Kelly Criterion is indeed the bet size optimisation parameter. However, it also shows the confidence level on a particular strategy, similar to the Sharpe ratio, the higher is generally better.

In a portfolio sense, it tells the user, how much to invest in this particular strategy as part of user's overall portfolio.

t2o2:
Is it possible to merge in?

kernc:
Sorry to have taken so long. CI was a mess. Thanks for the PR! :cake:

backtesting/_stats.py
```diff
@@ -137,6 +137,8 @@ def _round_timedelta(value, _period=_data_period(index)):
     s.loc['Profit Factor'] = returns[returns > 0].sum() / (abs(returns[returns < 0].sum()) or np.nan)  # noqa: E501
     s.loc['Expectancy [%]'] = returns.mean() * 100
     s.loc['SQN'] = np.sqrt(n_trades) * pl.mean() / (pl.std() or np.nan)
+    win_prob = (pl > 0).sum() / n_trades
+    s.loc['Kelly Criterion'] = win_prob - (1 - win_prob) / (pl[pl > 0].mean() / pl[pl < 0].mean())  # noqa: E501
```
kernc, line 141: @t2o2 Since `pl[pl < 0].mean()` is negative, don't we need to prefix it with a minus sign? :thinking:

t2o2: ChatGPT is right :)

How do you check it using ChatGPT? Do you just paste the function into the conversation line?

I will push the PR shortly.

t2o2: I can't managed to get ChatGPT to detect this negative problem, would be interesting to know how you did it. Thanks.

kernc: It took a few tries today after the recent update. But that's how I went about asking.
image
This is just one lucky results. I think after their recent update, it's even more obvious the machine doesn't have a clue what it's doing. 😆

PR #650, opened 2022-05-21
support named indicator

milliyang:
we can use named-indicator to make a goup;
so that,
 - two timeseries with different scale level can be viewed in different area
 - two timeseries with the scale level or relatived could be ploted together.

eg.
self.stddev   = self.I(SIGMA, self.data.Close, 30, TH.STDDEV,         overlay=False, named="stddev")
self.zscore   = self.I(SIGMA, self.data.Close, 30, TH.SIGMA,           overlay=False, named="zscore")
self.zpeak    = self.I(SIGMA, self.data.Close, 15, TH.SIGMA_PEAK, overlay=False, named="zscore")
self.zpeak    = self.I(SIGMA, self.data.Close, 30, TH.SIGMA_PEAK, overlay=False, named="zscore")

### DIFF
--- a/backtesting/_plotting.py
+++ b/backtesting/_plotting.py
@@ -518,6 +518,8 @@ return this.labels[index] || "";
         ohlc_colors = colorgen()
         indicator_figs = []
 
+        named_indicator = {}
+
         for i, value in enumerate(indicators):
             value = np.atleast_2d(value)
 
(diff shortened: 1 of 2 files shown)


kernc:
The preferred current approach to combine multiple indicators on one plot is:
```py
def init(self):
    def indicator():
        return (SIGMA(self.data.Close, 30, TH.SIGMA),
                SIGMA(self.data.Close, 15, TH.SIGMA_PEAK),
                SIGMA(self.data.Close, 30, TH.SIGMA_PEAK))  
    self.zscore = self.I(indicator, overlay=False)
    self.zpeak = self.zscore[2]
```

milliyang:
great !

PR #662, opened 2022-06-15
ENH: Add fixed commission option to _Broker (see #623)

### LINKED ISSUE #623: commission in dollar value
### Expected Behavior
able to set commission in dollars, disregard percentage. Some instruments I know (futures for example) has a fixed commission per trade, instead of percentage, there should be a switch to set % or $.

### Actual Behavior

only percentage is available

### Additional info

- Backtesting version: 0.3.3

mutt0-ds:
Hello and thanks for this amazing library! 
I added the 'use_fixed_commission' parameter to the class _Broker: if set to "True", in the function _adjusted_price the commission will be calculated as a flat fee instead of a percentage. 

Fixes #623

At first I thought about using parsing the value from commission (e.g. if the value is '0.05' -> percentage, if '1$' -> remove the dollar sign and parse it as fixed fee), but I don't want to mess up with the types so I added another parameter in the _Broker instance.  

What do you think? :)

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -749,7 +749,7 @@ class _Broker:
 
     def _adjusted_price(self, size=None, price=None) -> float:
         """
-        Long/shortprice, adjusted for commissions (fixed or percentage).
+        Long/short `price`, adjusted for commissions (fixed or percentage).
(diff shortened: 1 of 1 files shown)


mutt0-ds:
@kernc can you please let me know if I have to modify anything? Thank you!

PR #680, opened 2022-07-11
Make default n_jobs -1

andrii-solokh:


kernc:
See https://github.com/kernc/backtesting.py/discussions/276#discussioncomment-479430 for a brief discussion why `n_jobs=-1` is not the default for skopt optimization.

andrii-solokh:
@kernc Thank you for response. Very sad. I experience lack of performance, I would like to use all CPU cores and also GPU. But didn't find easy way for acceleration.

PR #720, opened 2022-08-17
Optionally display rich progress bar when backtesting or optimizing.

preritdas:
Some strategies are quite data/indicator heavy, especially when optimizing. I added optional `show_progress` boolean parameters to the `self.run` and `self.optimize` functions (default `False`) in their correct places, such that if they are passed in as `True` by the user, a visually-appealing Rich [progress bar](https://rich.readthedocs.io/en/stable/progress.html) is displayed in the console.

When backtesting, the progress bar represents the position within the dataset. When optimizing, the tracker wraps `_tqdm` in both the `'fork'` multiprocessing start method and the `'posix'` non-multiprocessing method.

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -1373,7 +1373,12 @@ class Backtest:
                                 for i in range(len(param_batches))]
 
                         # Optional Rich progress bar 
-                        if show_progress: futures_iter = track(as_completed(futures), "Optimizing...", len(futures))
+                        if show_progress: futures_iter = track(
+                            as_completed(futures), 
+                            "Grid optimizing with fork multiprocessing...", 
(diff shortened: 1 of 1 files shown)


preritdas:
I've updated it now so the progress bar is only displayed if multiprocessing is _off_, because we don't want multiple cores trying to simultaneously write their progress bars to the same console buffer. Instead, a neat status icon is shown. Below is an example of these stylistic features in action.

[asciicast](https://asciinema.org/a/3WahJXEjzVNOxvSYxhtKuri1u)

kernc:
I'm sorry for the expended effort, but I strongly prefer tqdm over Rich utils.

If there are other issues this PR addresses, please open a new PR for them.

Issue #149, opened 2020-09-25
Why don't use adjusted_price when close a trade ?


albertwh1te:
### Expected Behavior

I think the exit_price in trades should be the adjusted_price.
https://github.com/kernc/backtesting.py/blob/b456400e73c55afc66907013359b546b762050d4/backtesting/backtesting.py#L873
```Python
                    # Order size greater than this opposite-directed existing trade,
                    # so it will be closed completely
                    if abs(need_size) >= abs(trade.size):
                        # I think we shoud use adjusted_price 
                        self._close_trade(trade, adjusted_price, time_index)
                        need_size += trade.size
                    else:
                        # The existing trade is larger than the new order,
                        # so it will only be closed partially
                        self._reduce_trade(trade, adjusted_price, need_size, time_index)
                        need_size = 0
                        break

```
### Actual Behavior

We are using the price right now. :)  

### Steps to Reproduce

1. Just run a simple strategy
2. check the exit_price in stats._trades
3. compare with close price at the same

kernc:
The `adjusted_price` is only used when opening new trades:
https://github.com/kernc/backtesting.py/blob/b456400e73c55afc66907013359b546b762050d4/backtesting/backtesting.py#L887-L889
Thus, the spread/commissions are accounted for already at trade entry time (the broker ain't gonna wait for its share). If we were to use `adjusted_price` also when reducing/closing existing trades, the spread/commissions would be accounted for twice!

MarkWh1te:
> the spread/commissions would be accounted for twice!

In fact, it should be... for some exchange

> A client goes long 1 lot of USDCAD. The client is a ‘Silver’ account holder, so is charged $4 per $100,000 per side. 
> 
> Therefore, USD volume of 1 lot of USDCAD = $100,000
> 
> So, commission charged for this trade (open and close) = $4 x 2 (all commission charged and debited on the opening of the trade)
> 
> Total commission = $8
Link is [here ](https://www.houseofborse.com/commission-calculation)

kernc:
I see. :astonished: :sweat_smile: 

Disregarding the irrelevance of this broker, since this is a commission on currency conversion, I guess they are allowed to charge it twice. (_You convert EUR to GBP; here's our 0.004%. You convert GBP to EUR; here's our 0.004%._) It might be the same with some stock brokers. But, then again, this is certainly different from bid-ask spread charged by most spot brokers. :thinking: 

[...]

Issue #113, opened 2020-07-19
Minimum commission per trade


tankorsmash:
Say I've got a 1.00$ minimum commission + 1% per trade, it doesn't appear to be possible to account for that, beyond just using a the `commission=0.01` and ignoring the 1 dollar minimum.

Is such a thing possible?

### Additional info

- Backtesting version: 0.2.0

kernc:
Not possible. This would require an extra parameter, and we have nearly too many of those already, so I'm somewhat disinclined, more leaning to rule-of-thumb/ballpark estimations in that case.

Besides ignoring, I guess you can do accounting for (i.e. slightly increase commission rate for all trades, e.g. `commission=.0125`).

Do you make many trades where commissions sum to less than the minimum? In that case the relative commission model is completely flawed. :thinking:

casperbh96:
> Not possible. This would require an extra parameter, and we have nearly too many of those already, so I'm somewhat disinclined, more leaning to rule-of-thumb/ballpark estimations in that case.
> 
> Besides ignoring, I guess you can do accounting for (i.e. slightly increase commission rate for all trades, e.g. `commission=.0125`).
> 
> Do you make many trades where commissions sum to less than the minimum? In that case the relative commission model is completely flawed. But so might be your strategy ... 🤔

This issue is important for the whole repository. Most exchanges for stocks and futures charge a dollar commission, not a percentage like cryptocurrency. I think it's worth potentially adding a minimum commission nominated dollars/euros/etc. 

Also, consider that this adds flexibility and enables more accurate backtesting.
![image](https://user-images.githubusercontent.com/27340033/212732585-b58c0b91-b18d-4b84-bdbd-95195a71a9a7.png)

makovez:
commission is the commission ratio. E.g. if your broker's commission is 1% of trade value, set commission to 0.01. Note, if you wish to account for bid-ask spread, you can approximate doing so by increasing the commission, e.g. set it to 0.0002 for commission-less forex trading where the average spread is roughly 0.2‰ of asking price.

I have trouble to understand this phrase > Note, if you wish to account for bid-ask spread, you can approximate doing so by increasing the commission, e.g. set it to 0.0002 for commission-less forex trading where the average spread is roughly 0.2‰ of asking price.

if you set to 0.0002 then it would be 0.002% not 0.2% 

0.2% is 0.002% am i wrong?

kernc:
[...]

PR #839, opened 2022-12-13
ENH: Add Backtest(spread=), change Backtest(commission=)

kernc:
`commission=` is now applied twice as common with brokers. `spread=` takes the role `commission=` had previously.

Fixes https://github.com/kernc/backtesting.py/issues/149
Fixes https://github.com/kernc/backtesting.py/issues/113
Fixes https://github.com/kernc/backtesting.py/issues/1183
Closes https://github.com/kernc/backtesting.py/pull/662
Closes https://github.com/kernc/backtesting.py/issues/1006


casperbh96:
Looking forward to this one, it will enable this library to more accurately account for fees, especially with the commission tuple!

casperbh96:
@kernc when do you expect to merge this into the main branch and create a new release?

backtesting/backtesting.py
```diff
@@ -719,6 +731,9 @@ def __init__(self, *, data, cash, commission, margin,
         self.position = Position(self)
         self.closed_trades: List[Trade] = []
 
+    def _commission_func(self, order_size, price):
+        return self._commission_fixed + abs(order_size) * price * self._commission_relative
```
kernc, line 735: Will this cover all sorts of commission cases?

casperbh96: The base case is solved I would say.

**Nice to have:**
In cryptocurrency exchanges, your commissions tend to be reduced over a rolling 30-day period if you create enough volume.

It would be handy if you could create a scaling commission feature based on the transaction volume. It would also be handy if it can be split into maker/taker. 

image

jpvice:
Hi! First thanks @kernc thanks for building this project in the first place. I've been using it for a short while and it's awesome at what it does so far.

PR #887, opened 2023-01-23
Change to ubuntu-20.04

diegolovison:


kernc:
Switched to `ubuntu-latest` instead. Thanks for pointing it out! 😅

Issue #493, opened 2021-10-03
Broker Cash, Strange Metric


shaunpatterson:
### Expected Behavior
Start with $1000
Buy $100 of stock
self._broker._cash should be $900.  
Sell $150 of stock
self._broker._cash should be $1050

### Actual Behavior
Start with $1000
Buy $100 of stock
self._broker._cash doesn't change 
Sell $150 of stock
self._broker._cash is now $1050  (cash += closed_trade.pl)

I don't understand this behavior.  The variable is named cash, but it's not really the current cash value. 

If you want to make trades based on your actual available cash, self._broker.margin_available seems to be what you want... but that seems strange.

Am I missing something here?

kernc:
Can you rather show some code?

Note, everything prefixed with an underscore is, as customary, private API with no explicit guarantees.

shaunpatterson:
```
def _close_trade(self, trade: Trade, price: float, time_index: int):
        self.trades.remove(trade)
        if trade._sl_order:
            self.orders.remove(trade._sl_order)
        if trade._tp_order:
            self.orders.remove(trade._tp_order)

        self.closed_trades.append(trade._replace(exit_price=price, exit_bar=time_index))
        self._cash += trade.pl

    def _open_trade(self, price: float, size: int, sl: float, tp: float, time_index: int):
        trade = Trade(self, size, price, time_index)
        self.trades.append(trade)
        # Create SL/TP (bracket) orders.
        # Make sure SL order is created first so it gets adversarially processed before TP order
        # in case of an ambiguous tie (both hit within a single bar).
        # Note, sl/tp orders are inserted at the front of the list, thus order reversed.
        if tp:
            trade.tp = tp
        if sl:
            trade.sl = sl
```

Cash is only changed when the trade is closed.  

I understand the underscore variables are private -- however, I found this because I was interested in creating an enhancement to plot the available cash.

Issue #340, opened 2021-05-04
Plot the cash at the broker


gaborvecsei:
### Expected Behavior

We should have a flag for the `bt.plot()` function which would show the cash amount at our broker.

E.g.: `bt.plot(plot_broker_cash=True)`

Issue #124, opened 2020-08-03
Cash withdrawals and reinvestment


uberdeveloper:
I wanted to simulate some realistic scenarios relating to SIP, cash withdrawals, and partial reinvestment
Presently, I am manipulating the `self._broker._cash` attribute to this effect. 

Is there any suggested approach?
I am more concerned about the execution rather than the returns as I am using attributes inside the Strategy class to generate returns.

kernc:
Manipulating `self._broker._cash` seems reasonable for the time being.

Watch out for "margin call" liquidations if equity ever becomes negative:
https://github.com/kernc/backtesting.py/blob/28568d18c834ebd2d37b05794293ac6643691bab/backtesting/backtesting.py#L741-L743

uberdeveloper:
This is the present workaround and it works fine.

```python
class MyStrategy(Strategy):
    def init(self):
        # Anything above this capital in cash would be withdrawn
        self._initial_capital = 10000 
        # Everything else goes here

    def next(self):
        # After processing all logic
        
        # Withdraw only when there is excess
        if (self.equity > self._initial_capital) and (self._broker._cash > self._initial_capital):
            # Withdraw cash only when all positions are closed and everything is in cash
            if (self.position.size == 0):
                withdrawal = self._broker._cash - self._initial_capital
                self._broker._cash -= withdrawal
```

One big caveat is withdrawing cash when a position is already open. 
Would post after experimenting this

PR #925, opened 2023-03-02
Allow for cash to be deposited or withdrawn during a backtest

AdiSai:


kernc:
Sorry for the late turnaround. I'm not yet decided, but certainly not strongly opposed to this change. However, in the style of the rest of this library, I'd prefer `Strategy.cash` being a `@property`.

Implementing this would prevent people from accessing `self._broker._cash` as in https://github.com/kernc/backtesting.py/issues/493, https://github.com/kernc/backtesting.py/issues/340, https://github.com/kernc/backtesting.py/issues/124, ...

JamesonDiamond1 (approved):
@JamesonDiamond1

PR #941, opened 2023-03-21
Backtrader seems now abandoned

masafumimori:
I suppose Backtrader is now not maintained as its repo hasn't been updated since the last commit on Jul 18, 2021?
Maybe it is still usable but just thought update the doc here. (havent checked the others)
*I'm sorry if the package is still robust.


kernc:
🤷

PR #962, opened 2023-04-14
cash > 0

ElPettego:
more readable


david-pawlowski:
I don't think this is necessary, but I agree that this is more readable

PR #980, opened 2023-05-08
BUG: Allow multiple names for vector indicators (#382)

ivaigult:
Fixes https://github.com/kernc/backtesting.py/pull/382
Closes https://github.com/kernc/backtesting.py/pull/385

Previously we only allowed one name per vector indicator:

```
def _my_indicator(open, close):
    return tuple(
        _my_indicator_one(open, close),
         _my_indicator_two(open, close),
    )

self.I(
    _my_indicator,
    # One name is used to describe two values
    name="My Indicator",
    self.data.Open,
    self.data.Close
)
```

Now, the user can supply two (or more) names to annotate each value individually. The names will be shown in the plot legend. The following is now valid:

```
self.I(
    _my_indicator,
    name=["My Indicator One", "My Indicator Two"],
    self.data.Open,
    self.data.Close
)
```


backtesting/backtesting.py
```diff
@@ -115,13 +117,21 @@ def I(self,  # noqa: E743
...
             name = (f'{func_name}({params})' if params else f'{func_name}')
+        elif isinstance(name, str):
+            name = _format_name(name)
+        elif try_(lambda: all(isinstance(item, str) for item in name), False):
+            name = [_format_name(item) for item in name]
         else:
-            name = name.format(*map(_as_str, args),
-                               **dict(zip(kwargs.keys(), map(_as_str, kwargs.values()))))
+            raise TypeError(f'Unexpected `name` type {type(name)}, `str` or `Iterable[str]` '
+                            'was expected.')
```
kernc, line 134:
```suggestion
            raise TypeError(f'Unexpected `name=` type {type(name)}; expected `str` or `Sequence[str]`')
```
`Iterable` such as a generator would get expent by our `isinstance()` checks.

backtesting/backtesting.py
```diff
@@ -90,7 +90,9 @@ def I(self,  # noqa: E743
         same length as `backtesting.backtesting.Strategy.data`.
 
         In the plot legend, the indicator is labeled with
-        function name, unless `name` overrides it.
+        function name, unless `name` overrides it. If `func` returns
+        multiple arrays, `name` can be a collection of strings, and
+        the size must agree with the number of arrays returned.
```
kernc, line 95:
```suggestion
        multiple arrays, `name` can be a sequence of strings, and
        its size must agree with the number of arrays returned.
```
`set` is a standard "collection", yet its ordering is undefined (not even dict-like).

backtesting/_plotting.py
```diff
@@ -537,10 +537,24 @@ def __eq__(self, other):
...
             colors = colors and cycle(_as_list(colors)) or (
                 cycle([next(ohlc_colors)]) if is_overlay else colorgen())
-            legend_label = LegendStr(value.name)
-            for j, arr in enumerate(value, 1):
+
+            tooltip_label = value.name
+
+            if len(value) == 1:
+                assert isinstance(value.name, str)
+                legend_labels = [LegendStr(value.name)]
```
kernc, line 545: This case can be handled by:
```py
legend_labels = [LegendStr(item) for item in _as_list(value.name)]
```
in the branch below. This will then also support use like:
```py
self.I(indicator_1d, name=['foo'])
```
which is reasonable.

kernc (approved):
Indeed, this looks good and is properly classified (bug). 😆

ivaigult:
Hi @kernc,

Thank you for the feedback! I think I addressed everything. Could you, please, resolve all conversations if it looks okay to you.

Thank you:slightly_smiling_face:

backtesting/_plotting.py
```diff
@@ -537,10 +537,22 @@ def __eq__(self, other):
...
+
+            if len(value) == 1:
+                legend_labels = [LegendStr(item) for item in _as_list(value.name)]
+            elif isinstance(value.name, str):
+                legend_labels = [
+                    LegendStr(f"{name}[{index}]")
+                    for index, name in enumerate(repeat(value.name, len(value)))
+                ]
+            else:
+                legend_labels = [LegendStr(item) for item in value.name]
```
kernc, line 551:
```suggestion
            if isinstance(value.name, str):
                legend_labels = [
                    LegendStr(f"{name}[{index}]")
                    for index, name in enumerate(repeat(value.name, len(value)))
                ]
            else:
                legend_labels = [LegendStr(item) for item in _as_list(value.name)]
```

ivaigult: Hmm, this does not seem to be correct :thinking: 

There are three use cases:
 * The indicator returns one value, one name `name` is passed. The label will be `name`.
 * The indicator returns multiple values, one name `name` is passed. The labels will be `name[0]`, `name[1]`, ...
 * The indicator returns multiple values, multiple names `name_one`, `name_two`, ... are passed. The labels will be the same as names.

There are three branches exactly to account for each use case. Now, if we merge this suggestion, the first use case will be broken, as the label will be `name[0]`. Am I missing something?

ivaigult: I think I understand your point: there are two very similar branches. I got rid of one of them, and my scenarios above still work. Please, have a look.

backtesting/_plotting.py
```diff
@@ -537,10 +537,20 @@ def __eq__(self, other):
...
+
+            tooltip_label = value.name if isinstance(value.name, str) else ", ".join(value.name)
+
+            if isinstance(value.name, str) and len(value) > 1:
+                legend_labels = [
+                    LegendStr(f"{name}[{index}]")
+                    for index, name in enumerate(repeat(value.name, len(value)))
+                ]
+            else:
+                legend_labels = [LegendStr(item) for item in _as_list(value.name)]
```
kernc, line 549: > the first use case will be broken, as the label will be `name[0]`

Absolutely, you're right. I didn't follow it through correctly. 😅 With no such consolidations applicable, I'd therefore prefer it simply top-down as we reason about it.

```suggestion
            if isinstance(value.name, str):
                tooltip_label = value.name
                if len(value) == 1:
                    legend_labels = [LegendStr(value.name)]
                else:
                    legend_labels = [
                        LegendStr(f"{value.name}[{i}]")
                        for i in enumerate(len(value))
                    ]
            else:
                tooltip_label = ", ".join(value.name)
                legend_labels = [LegendStr(item) for item in value.name]
```

backtesting/_plotting.py
```diff
@@ -3,7 +3,7 @@
 import sys
 import warnings
 from colorsys import hls_to_rgb, rgb_to_hls
-from itertools import cycle, combinations
+from itertools import cycle, combinations, repeat
```
kernc, line 6:
```suggestion
from itertools import cycle, combinations
```

kernc:
> gtg

It is. The checks won't finish near green, though. Project docs build hinges on scikit-optimize being presently unmaintained and in a defunct state wrt other Python/numpy/sklearn stack. I'm not yet certain what to do about that.

romanbsd:
include the scikit-optimize, you actually need just a few files as far as I remember.

ivaigult:
Are there any obstacles that preventing merge of this?

> scikit-optimize being presently unmaintained and in a defunct state

I understand it's an issue, but I appreciate this will be addressed separately in another PR?

PR #200, opened 2020-12-17
Add Order.tag for tracking orders and trades

qacollective:
Fixes https://github.com/kernc/backtesting.py/issues/197
Fixes https://github.com/kernc/backtesting.py/issues/53

Hello again.  Many apologies for my butchering of the PR process.  Git is certainly something I am no expert at.  :cry: 

This is another PR which implements the tagging feature, makes tags an object type and fixes previous pep8 issues.

I am currently testing this for the first time on my own trading platform.  The code seems to work fine to persist tags at the moment, but there are far too many trades being taken out, so unless that's a likely outcome of these changes, that problem is likely with my code outside of backtesting.py.

PR #995, opened 2023-05-28
BUG: trying to round none / string(tag) / Boolean (is_contingent)

ajwinkworth:
BUG: Fix attempting to round none / string(tag) / Boolean (is_contingent) in backtesting.py:408

Optional float values(limit, stop, sl, tp) can be float | None. Testing for none to not pass it to round()
Boolean values (is_contingent) shouldnt be rounded.
Str values (tag) shouldn't be rounded.

See:
[A fix for Comment](https://github.com/kernc/backtesting.py/pull/200#issuecomment-1446897531)

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -405,17 +405,19 @@ class Order:
             setattr(self, f'_{self.__class__.__qualname__}__{k}', v)
         return self
 
+    
     def __repr__(self):
-        return '<Order {}>'.format(', '.join(f'{param}={round(value, 5)}'
+        return '<Order {}>'.format(', '.join(f'{param}={value}'
                                              for param, value in (
-                                                 ('size', self.__size),
-                                                 ('limit', self.__limit_price),
-                                                 ('stop', self.__stop_price),
-                                                 ('sl', self.__sl_price),
-                                                 ('tp', self.__tp_price),
+                                                 ('size', round(self.__size,5)),
+                                                 ('limit', round(self.__limit_price,5) if self.__limit_price is not None else None),
+                                                 ('stop', round(self.__stop_price,5) if self.__stop_price is not None else None),
+                                                 ('sl', round(self.__sl_price,5) if self.__sl_price is not None else None),
+                                                 ('tp', round(self.__tp_price,0) if self.__tp_price is not None else None),
                                                  ('contingent', self.is_contingent),
                                                  ('tag', self.__tag),
                                              ) if value is not None))
+
 
     def cancel(self):
         """Cancel the order."""


backtesting/backtesting.py
```diff
@@ -405,18 +405,20 @@ def _replace(self, **kwargs):
             setattr(self, f'_{self.__class__.__qualname__}__{k}', v)
         return self
 
+
```
kernc, line 408: [PEP 8 prescribes](https://peps.python.org/pep-0008/#blank-lines) **one empty line** between methods.

backtesting/backtesting.py
```diff
@@ -405,18 +405,20 @@ def _replace(self, **kwargs):
...
-                                                 ('size', self.__size),
-                                                 ('limit', self.__limit_price),
-                                                 ('stop', self.__stop_price),
-                                                 ('sl', self.__sl_price),
-                                                 ('tp', self.__tp_price),
+                                                 ('size', round(self.__size,5)),
+                                                 ('limit', round(self.__limit_price,5) if self.__limit_price is not None else None),
+                                                 ('stop', round(self.__stop_price,5) if self.__stop_price is not None else None),
+                                                 ('sl', round(self.__sl_price,5) if self.__sl_price is not None else None),
+                                                 ('tp', round(self.__tp_price,0) if self.__tp_price is not None else None),
```
kernc, line 416: I'd prefer this `round(value, 5) if value is not None else ''` be done once in the comprehension value expression (as before).

Also don't see a problem with `round(is_contingent, 5)` since the boolean rounds to either `1` or `0`.

kernc (changes requested):

backtesting/backtesting.py
ajwinkworth: How would you prefer to deal with string tags then? Take them out of the comprehension and then append them separately?

ajwinkworth: After some further investigation it could be:
```python
    def __repr__(self):
        return '<Order {}>'.format(', '.join(f'{param}={round(value, 5) if isinstance(value, SupportsRound) else value}'
                                             for param, value in (
                                                 ('size', self.__size),
                                                 ('limit', self.__limit_price),
                                                 ('stop', self.__stop_price),
                                                 ('sl', self.__sl_price),
                                                 ('tp', self.__tp_price),
                                                 ('contingent', self.is_contingent),
                                                 ('tag', self.__tag),
                                             ) if value is not None))
```

Which I haven't tested yet. Would just need an additional import from typing for SupportsRound.

kernc: Something like that. Can also use [`isinstance(value, Number)`](https://docs.python.org/3/library/numbers.html#numbers.Number) like we do in [some other places](https://github.com/search?q=repo%3Akernc%2Fbacktesting.py+isinstance+number&type=code).

jaysonfrancis:
```python
  def __repr__(self):
      return '<Order {}>'.format(', '.join(f'{param}={round(value, 5) if isinstance(value, (int, float)) else value}'
                                           for param, value in (
                                               ('size', self.__size),
                                               ('limit', self.__limit_price),
                                               ('stop', self.__stop_price),
                                               ('sl', self.__sl_price),
                                               ('tp', self.__tp_price),
                                               ('contingent', self.is_contingent),
                                               ('tag', self.__tag),
                                           ) if value is not None))
```

kernc:
Thanks and apologies. Already fixed in https://github.com/kernc/backtesting.py/commit/8dd1e3696f0453f1276c4ab10f358c18b72a61c9.

PR #1028, opened 2023-07-22
Added `tqdm` to  setup.py

abd1rahmane:
If 'tqdm' is not installed, no warning will show and the progress will not be displayed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ if __name__ == '__main__':
                 'matplotlib',
                 'scikit-learn <= 1.1.3',  # Pinned due to boken scikit-optimize
                 'scikit-optimize',
-                'tqdm'
+                'tqdm',
             ],
             'dev': [
                 'ruff==0.0.160',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -48,6 +48,7 @@
                 'matplotlib',
                 'scikit-learn <= 1.1.3',  # Pinned due to boken scikit-optimize
                 'scikit-optimize',
+                'tqdm'
```
david-pawlowski, line 51: Add comma at the end of list items as good practice. It won't show that this line was changed(comma added) after someone add new item.

abd1rahmane: I see, thanks.

kernc:
Thanks. And apologies. That's the intended behavior—tqdm is wholly optional. We can discuss the warning ... 🤷
https://github.com/kernc/backtesting.py/blob/eefff8792872034fae312b795fa4853ec7cd0561/backtesting/backtesting.py#L28-L33

PR #1037, opened 2023-08-04
Update lib.py to fix #901

### LINKED ISSUE #901: Error of plot_heatmaps()
### Expected Behavior
Fix the mistake or make it a normal explanation.

### Actual Behavior
```python
Traceback (most recent call last):
  File "... .py", line 298, in optimize
    plot_heatmaps(heatmap, agg='mean')
  File "... /backtesting/lib.py", line 141, in plot_heatmaps
    return _plot_heatmaps(heatmap, agg, ncols, filename, plot_width, open_browser)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "... /backtesting/_plotting.py", line 685, in plot_heatmaps
    low=min(df.min().min() for df in dfs),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: min() arg is an empty sequence
```
There `dfs` specified in the error is an empty [].

### Steps to Reproduce
```python
from backtesting.lib import plot_heatmaps

...
stats, heatmap = bt.optimize(a_test_param=range(5), maximize='Equity Final [$]', return_heatmap=True)

plot_heatmaps(heatmap, agg='mean')
```
The error is raised when `plot_heatmaps()` run. 
I think, the reason is that for a graph with two axes (X, Y) need at least 2 optimized variables. Therefore, this method does not accept the `heatmap` that is derived from `bt.optimize()` that was run with only one parameter to be optimized.

### Additional info

- Backtesting version: 0.3.3

siufai6:
Fixes https://github.com/kernc/backtesting.py/issues/901

Heatmap plots requires at least two parameters (variables) in the heatmap Series. Add an assert before calling _plot_heatmap to guard against the case of having only 1 variable.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ first [fork the project]. Then:
 
     git clone git@github.com:YOUR_USERNAME/backtesting.py
     cd backtesting.py
-    pip3 install -e .[doc,test,dev]
+    pip3 install -e '.[doc,test,dev]'
 
 [fork the project]: https://help.github.com/articles/fork-a-repo/
 


kernc:
Thanks. And apologies—already fixed in https://github.com/kernc/backtesting.py/commit/cb58822bb182337b3a997df1096a446fb5dc7167.

PR #1039, opened 2023-08-06
Added TP and SL into stats file.

aligheshlaghi97:
When an algorithmic trader seeks to analyze a trading strategy, the Take Profit (TP) and Stop Loss (SL) values emerge as pivotal considerations within that strategy. Thus, an inclusion of these significant values into the stats file becomes paramount.


kjpou1:
Agree.  I ended up using the `.tag` and custom `StrategyTrade` class to accumulate all this information.

PR #1054, opened 2023-08-19
added show_output arg to backtesting/backtesting.py plot method

hadif1999:


kernc:
Apologies, how the the new behavior, provided by this PR, differ from simply passing [`bt.plot(..., open_browser=False)`](https://kernc.github.io/backtesting.py/doc/backtesting/backtesting.html#backtesting.backtesting.Backtest.plot)?

PR #1102, opened 2024-01-02
allow open trades at end of backtest.

themantalope:
Added option to leave trades open at the end of a backtest. Allows users to check if they have open trades at the end of a backtest as an easy way to use the strategy in a "live" manner.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ doc/build/*
 *~*
 
 .venv/
+build/*
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -1074,9 +1074,9 @@ class Backtest:
 
         If `exclusive_orders` is `True`, each new order auto-closes the previous
         trade/position, making at most a single trade (long or short) in effect
-        at each time.
+        at each time.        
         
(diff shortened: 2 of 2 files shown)


LorenzoPietroBonatti:
Hey, could you please show me how is it working for you? I tried to fork your branch and use it, but the trades doesn't remained open. Even if my strategy should be bought in the past candles and should not have sold all my position. Appreciate the help.

themantalope:
Sure, here is a quick example:

```python
from other_module import AmazingStrategy
from backtesting import Backtest
import pandas as pd

data = pd.read_csv("filewith_olhc.csv")
bt = Bactest(data, AmazingStrategy, cash=10000, leave_orders_open=True)
results = bt.run()
# now get the strategy object from results
strat_res = results._strategy
open_trades = strat_res.trades
```

Typically, all trades are closed at the end of the `Backtest.run` (see [here](https://github.com/kernc/backtesting.py/blob/0ce24d80b1bcb8120d95d31dc3bb351b1052a27d/backtesting/backtesting.py#L1222)). I wanted to be able to take a strategy with updated daily data and see if there are any "open" trades/positions (not sure which is the better term to use). If there is a less clunky way to do so please let me know.

Issue #39, opened 2020-01-18
How to add custom data


codingalex:
I would like to use other data such as sentiment for my backtests. Is there any way to custom data to self.data?

kernc:
Whatever columns you pass in in the initial dataframe should remain available. From the [docs](https://kernc.github.io/backtesting.py/doc/examples/Quick%20Start%20User%20Guide.html#Data):
> Your data frames can have other columns

Is this what you're looking for?

codingalex:
That's true! Thank you so much :)

PR #1111, opened 2024-01-26
Iteration with Single-Step Progression to Improve Flexibility and Adaptability in Algorithmic Strategies

IperGiove:
It is possible to split the run() function into two distinct functions, allowing for a single-step iteration on the ticks:
- initialize(): Saves the data and the strategy into a self variable.
- next(): Moves to the next tick in the iteration.

In which scenario do you want to use a next() function instead of a run()?
Using a next() function allows passing any kwargs to the Strategy.next() function. This flexibility enables processing additional information outside the strategy class and injecting it as needed.
However the run function is preserved in order to have the back-compatibility.

For example, one can utilize backtesting.py as an engine to train a reinforcement learning model, requiring an observation and an action at each step.

I've included a test/_iteration.py  with a custom TestStrategy that works with both run and iteration, demonstrating how you can inject any parameters into the next function.
Here an example how I use the single step iteration in order to training an Env with gym and stable-baseline3 library:
https://gist.github.com/IperGiove/325832e30df44639ec78a618b84daf3c

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -1662,3 +1662,85 @@ class Backtest:
             reverse_indicators=reverse_indicators,
             show_legend=show_legend,
             open_browser=open_browser)
+
+    def initialize(self, **kwargs):
+        """
+        Initialize the backtest with the given parameters. Keyword arguments are interpreted as strategy parameters.
+
+        :param kwargs: Strategy parameters
+        """
+        data = _Data(self._data.copy(deep=False))
+        broker: _Broker = self._broker(data=data)
+        strategy: Strategy = self._strategy(broker, data, kwargs)
+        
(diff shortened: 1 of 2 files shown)


JamesonDiamond1 (approved):
@JamesonDiamond1

kernc:
Ok, I think I understand the issue. You can [pass custom data to the strategy logic (df columns in addition to OHLC)](https://github.com/kernc/backtesting.py/issues/39), but you can't pass custom objects (your RL optimizer object that tells your next action). In that case, you can:
```py
df['model_obj'] = np.repeat(rl_model_obj, len(df))
```
and in `next()` use that (i.e. `self.data.model_obj[-1]`).

Would that work for you?

PR #1115, opened 2024-02-11
Fix rounding error when tag is not numeric.

kjpou1:


kjpou1:
Should have been local PR

PR #1119, opened 2024-02-27
Added Order Fees

RP1-MVac:
Some brokers don't ask for a fixed commission but instead want a fee for every trade conducted. You can now add an order_fees parameter to the backtesting instance and the fees will show up in a new entry in the result.

if order_fees are not assigned nothing changes

if order_fees are assigned the final report will look like this:
...
Equity Final [$] (without fees)            30228.33435
Equity Final [$]                           30172.33435
...


kernc:
This is now possible with fixed or variable commission since https://github.com/kernc/backtesting.py/pull/839. Soon to be released.

PR #1131, opened 2024-03-28
feat: added color bar for each heatmap

devon-n:
Moved the cmap from plot_heatmaps to within the for loop that creates heatmaps so that each heatmap can have its own colorbar.

Placed the colorbar on the right of each heatmap


JamesonDiamond1 (approved):
@JamesonDiamond1

PR #1164, opened 2024-08-06
Rounding error in RGB functionality

Murat-U-Saglam:
The values before passed to Bokeh were not rounded as integers and were treated as float due to a decimal point. I have added robust checking and enforcement of ints within that function.

Bokeh RGB accepts integers, this error causes issues with serialisation as discussed in  [Issue](https://github.com/bokeh/bokeh/issues/14011#event-13771094080)

### DIFF
--- a/backtesting/_plotting.py
+++ b/backtesting/_plotting.py
@@ -88,7 +88,8 @@ def colorgen():
 def lightness(color, lightness=.94):
     rgb = np.array([color.r, color.g, color.b]) / 255
     h, _, s = rgb_to_hls(*rgb)
-    rgb = np.array(hls_to_rgb(h, lightness, s)) * 255.
+    rgb = np.array(hls_to_rgb(h, lightness, s)) * 255
+    rgb = np.rint(rgb)
     return RGB(*rgb)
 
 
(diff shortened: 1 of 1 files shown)


bryevdv:
FYI please check this works, I am not sure calling `rint` will be sufficient as it still (AFAIK) generates floating point values, that just happen to be integer values. But these may render like `2.`, `10.` etc. and I believe this will still trip up the regex based input validation. Though really I would recommend collecting the results in a tuple and actually calling `int` on all the components to be safest.

backtesting/_plotting.py
```diff
@@ -88,7 +88,8 @@ def colorgen():
 def lightness(color, lightness=.94):
     rgb = np.array([color.r, color.g, color.b]) / 255
     h, _, s = rgb_to_hls(*rgb)
-    rgb = np.array(hls_to_rgb(h, lightness, s)) * 255.
+    rgb = np.array(hls_to_rgb(h, lightness, s)) * 255
+    rgb = np.rint(rgb)
```
kernc, line 92:
```suggestion
    rgb = (np.array(hls_to_rgb(h, lightness, s)) * 255).astype(int)
```

kernc:
Hope this works. Thanks!

PR #1176, opened 2024-10-05
Adding TQDM progress bar to Run method

vineetvermait:
Adding TQDM progress bar to Run method.

This will start showing progress while executing the backtest


kernc:
Nice, but. Why not wrap the `range` call with the existing `_tqdm()` function?
The [case seems supported](https://tqdm.github.io/#:~:text=for%20i%20in-,tqdm(range(10000))).

PR #1177, opened 2024-10-09
fixing pandas read_csv

ljtron:


kernc:
Fixed in master. Thanks! 😊

Issue #88, opened 2020-06-15
bt.plot possible enhancement - plot organization and/or grouping. 


sidsidsidsid:
Strategy1 below:
I noticed that an overlayed plot would be created for self.m1, self.m2, and self.m3, and candlestick. But, 3 plots will be created for each self.m4, self.m5, and self.m6. Is there a way to overlay a second plot that would combine self.m4, self.m5, and self.m6?

Strategy2 below:
If the above is not possible, I have an enhancement request. Add a flag to selfI() that tells how the plots should be organized. A flag would tell the grouping/overlays and the order of the plots. See below

```py
class Strategy1(Strategy):
    def init(self):
        Close = self.data.Close
        self.m1  = self.I(Indicator_1, Close)
        self.m2  = self.I(Indicator_1, Close)
        self.m3  = self.I(Indicator_1, Close)
       
	Var = (self.data.High + self.data.Close) * 100
        
        self.m4  = self.I(Indicator_2, Var)
        self.m5  = self.I(Indicator_2, Var)
        self.m6  = self.I(Indicator_2, Var)

     def next(self):
        ...

class Strategy2(Strategy):
    def init(self):
        Close = self.data.Close
        self.m1  = self.I(Indicator_1, Close,overlay_group=1)
        self.m2  = self.I(Indicator_1, Close,overlay_group=1)
        self.m3  = self.I(Indicator_1, Close,overlay_group=1)
       
	Var = (self.data.High + self.data.Close) * 100
       
        self.m4  = self.I(Indicator_2, Var,overlay_group=2)
        self.m5  = self.I(Indicator_2, Var,overlay_group=2)
        self.m6  = self.I(Indicator_2, Var,overlay_group=2)

     def next(self):
        ...
```

sidsidsidsid:
Typo  -->   class Strategy@ = class Strategy2

kernc:
`self.I(..., overlay=True)`, [as documented](https://kernc.github.io/backtesting.py/doc/backtesting/backtesting.html#backtesting.backtesting.Strategy.I), already forces the defined indicator to be overlaid on the OHLC chart. By default, a heuristic determines whether the indicator should be overlaid or not.

When an indicator is not overlaid over OHLC chart, it is plotted separately under it.

One way to plot multiple non-overlaid indicators on the same underneath plot is to define a new indicator that returns a _tuple of indicators_ that are to be plotted together: 
```py
    def init(self):
        def new_indicator(values):
[...]

PR #1217, opened 2025-02-12
Adding plot_group argument to indicator constructor

vlasenkoalexey:
This would allow combining few indicators in one plot.


kernc:
Have you tried returning multiple arrays in your indicator func like in https://github.com/kernc/backtesting.py/issues/88#issuecomment-645336450 ? Maybe the docs could be clearer.

PR #1221, opened 2025-02-16
added beta & alpha

jensnesten:
Add Alpha and Beta Metrics to Performance Analysis

This PR introduces two key risk-adjusted performance metrics:
- Alpha: Measures strategy's excess return relative to buy & hold benchmark
- Beta: Measures strategy's systematic risk (correlation with market returns)


backtesting/_stats.py
```diff
@@ -131,6 +131,20 @@ def _round_timedelta(value, _period=_data_period(index)):
...
+    market_returns = []
+    # Calculate returns for each period
+    for i in range(1, len(equity)):
+        equity_return = (equity[i] - equity[i - 1]) / equity[i - 1]
+        market_return = (c[i] - c[i - 1]) / c[i - 1]
+        equity_returns.append(equity_return)
+        market_returns.append(market_return)
+    
+    equity_returns = np.array(equity_returns)
+    market_returns = np.array(market_returns)
```
kernc, line 145: Could this be:
```py
equity_returns = np.diff(equity) / equity[:-1]
market_returns = np.diff(c) / c[:-1]
```
or
```py
equity_returns = np.diff(np.log(equity))
market_returns = np.diff(np.log(c))
```
or something simple and vectorized?

Make it short enough, and you can move it right where it's used. 🤓

jensnesten: sure, I suppose you mean like:

```python
    equity_returns = np.diff(equity) / equity[:-1]
    market_returns = np.diff(c) / c[:-1]
    cov_matrix = np.cov(equity_returns, market_returns)
    s.loc['Beta'] = round(cov_matrix[0, 1] / cov_matrix[1, 1], 2)
```
where the calc are down by the rest of the stat definitions.

backtesting/_stats.py
```diff
@@ -151,6 +151,12 @@ def _round_timedelta(value, _period=_data_period(index)):
     with np.errstate(divide='ignore'):
         s.loc['Sortino Ratio'] = (annualized_return - risk_free_rate) / (np.sqrt(np.mean(day_returns.clip(-np.inf, 0)**2)) * np.sqrt(annual_trading_days))  # noqa: E501
     max_dd = -np.nan_to_num(dd.max())
+    s.loc['Alpha [%]'] = s.loc['Return [%]'] - s.loc['Buy & Hold Return [%]']
+    # calculate returns using vectorized operations
+    equity_returns = np.diff(equity) / equity[:-1]
+    market_returns = np.diff(c) / c[:-1]
```
kernc, line 157: Is there a reason to prefer simple instead of log-returns? This was also suggested by the LLM.

kernc (changes requested):

backtesting/_stats.py
jensnesten: No, I just tend to test on high frequency data on relatively short periods, so the difference between simple and log would be negligible. It'll be more precise with log returns, you're right - I've changed it.

backtesting/_stats.py
```diff
@@ -151,6 +151,12 @@ def _round_timedelta(value, _period=_data_period(index)):
...
+    cov_matrix = np.cov(equity_returns, market_returns)
+    s.loc['Beta'] = round(cov_matrix[0, 1] / cov_matrix[1, 1], 2)
     s.loc['Calmar Ratio'] = annualized_return / (-max_dd or np.nan)
```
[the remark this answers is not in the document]
jensnesten: Sure! ahh ok, my mistake. I'll leave out the round()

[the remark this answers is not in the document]
jensnesten: had to move beta above alpha to use it in calculation

backtesting/_stats.py
```diff
@@ -151,6 +151,12 @@ def _round_timedelta(value, _period=_data_period(index)):
     with np.errstate(divide='ignore'):
         s.loc['Sortino Ratio'] = (annualized_return - risk_free_rate) / (np.sqrt(np.mean(day_returns.clip(-np.inf, 0)**2)) * np.sqrt(annual_trading_days))  # noqa: E501
     max_dd = -np.nan_to_num(dd.max())
+    s.loc['Alpha [%]'] = s.loc['Return [%]'] - s.loc['Buy & Hold Return [%]']
```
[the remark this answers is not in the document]
jensnesten: No you're right, i've changed it to:
```python
    s.loc['Alpha [%]'] = (s.loc['Return [%]'] - risk_free_rate * 100) - s.loc['Beta'] * (s.loc['Buy & Hold Return [%]'] - risk_free_rate * 100)
```

kernc (commented):
This is looking great!

You'll need to amend [a few tests](https://github.com/kernc/backtesting.py/actions/runs/13485433836/job/37728976524) to get the lights green ...

backtesting/_stats.py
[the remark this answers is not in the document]
kernc: Are the values in both cases comparable, did you happen to check?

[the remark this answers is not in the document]
jensnesten: yes I checked everything before pushing, and the results are comparable.

jensnesten:
When looking at the expected stats results, mine are quite off. Running the smacross strat as defined in _test.py i get the following output:

```bash
Start                     2004-08-19 00:00:00
End                       2013-03-01 00:00:00
Duration                   3116 days 00:00:00
Exposure Time [%]                    93.99441
Equity Final [$]                  39770.03122
Equity Peak [$]                     61303.036
Commissions [$]                    8584.81878
Return [%]                          297.70031
Buy & Hold Return [%]               522.06019
Return (Ann.) [%]                    17.58149
Volatility (Ann.) [%]                35.35528
CAGR [%]                             11.81185
Sharpe Ratio                          0.49728
Sortino Ratio                         0.89695
Calmar Ratio                          0.34926
Alpha [%]                           277.50897
Beta                                  0.03868
Max. Drawdown [%]                   -50.33958
Avg. Drawdown [%]                    -6.25468
Max. Drawdown Duration      963 days 00:00:00
Avg. Drawdown Duration       47 days 00:00:00
# Trades                                   65
Win Rate [%]                         46.15385
Best Trade [%]                       53.59595
Worst Trade [%]                     -18.39887
Avg. Trade [%]                        2.35371
Max. Trade Duration         183 days 00:00:00
Avg. Trade Duration          46 days 00:00:00
Profit Factor                         2.08802
Expectancy [%]                        3.09763
SQN                                   1.01383
Kelly Criterion                       0.14365

```

These do not line up with the expected output from the dataframe defined in test_compute_stats(). Seems like the last trade isn't closed properly, it should be 66 trades in total. Ran a simple strat that buys on first close and sells at the last, and it failed to close the trade as well (which works fine on my older version of this repo). Also the test results i get from this newer version differ from my older version when running the same strat, can't seem to tell why though - just though I'd mention.

kernc:
> These do not line up with the expected output from the dataframe defined in test_compute_stats().

Your stats contain `'Commissions [$]'` but no `test_compute_stats` specifies them ... 💡
https://github.com/kernc/backtesting.py/blob/ad6d6e0bbdb6114f7cf927bc9a543c07f499e2e3/backtesting/test/_test.py#L282-L283
https://github.com/kernc/backtesting.py/blob/ad6d6e0bbdb6114f7cf927bc9a543c07f499e2e3/backtesting/test/_test.py#L902-L903

There were a few changes recently that might change results obtained with prior versions, notably:

https://github.com/kernc/backtesting.py/commit/ae3d69fa0abb765da0b11ba19ee1b29f894e4e94*
https://github.com/kernc/backtesting.py/commit/d4ec0baba7f9bafd8dd0cdcb317d46e3d3a09003
https://github.com/kernc/backtesting.py/commit/8fbb902a8cf45280078745dbb02a7e3ae01f2293
https://github.com/kernc/backtesting.py/commit/8b88e8195526ec10203bb0830583e78200bc56c8
https://github.com/kernc/backtesting.py/commit/58de404db50d0675c5d96780e48b20993cc6d783

The current tests are considered accurate as long as they are [passing on CI](https://github.com/kernc/backtesting.py/actions/runs/13468864068).

jensnesten:
Ahhh I see now, thank you! Have added some expected values of alpha and beta to the dataframe. Is that all that is needed?

.gitignore
```diff
@@ -22,3 +22,6 @@ build/*
 *~*
 
 .venv/
+
+.DS_Store
+**/.DS_Store
```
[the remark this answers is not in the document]
jensnesten: You are so right, my apologies!

README.md
```diff
@@ -67,6 +67,8 @@ CAGR [%]                                16.80
 Sharpe Ratio                             0.66
 Sortino Ratio                            1.30
 Calmar Ratio                             0.77
+Alpha [%]                              450.62
+Beta                                     0.02
```
kernc, line 71: For given values of `s.loc['Return [%]']` and `s.loc['Buy & Hold Return [%]']`, the previous computation of
```py
s.loc['Alpha [%]'] = s.loc['Return [%]'] - s.loc['Buy & Hold Return [%]']  # == -144
```
seemed much more reasonable! Can you explain the discrepancy?

kernc: LLM suggests, among the less applicable options, that the values are not computed on the same time scale—beta is computed on daily returns and alpha on the whole duration ... 🤔

jensnesten: So the 'simple' alpha we first had is absolute alpha, where we simply calculate the absolute difference of our strategy relative to the underlying benchmark. The new calc is risk-adjusted alpha, which is the official definition of alpha, but here we might end up in situations where alpha is positive, despite underperforming the benchmark in terms of return. Especially in beta neutral strategies where beta approaches 0, you'd have situations where the right-most term will go to 0 - essentially signaling very little risk:

$`\alpha=(R-r) - \beta(R_{b}-r)`$
when $`\beta \rightarrow 0`$:
$`\alpha=(R-r)`$

I personally like the simple alpha calc in this context, because its simple and intuitive. But the alpha we have now is the correct alpha definition.

kernc: Thanks. I added a source code comment so that we remain aware of this counter-intuitive issue. Hope not too many users get confused by it, or that we eventually get to improve on it.

kernc (approved):
Looks good! Thanks for working through it!

Issue #963, opened 2023-04-15
The atr introduce look ahead bias or is my interpretation incorrect?


EladioRocha:
### Expected Behavior
When we trade with trailing stop loss the ATR is calculated. Fo example suppose that we calculate the ATR for 100 periods the first 100 periods should be removed because they don't have values and if you try to open a trade in the period 10 there isn't an available ATR value therefore isn't possible to determine the stop loss. So in that case all the values that contains NaN values should be removed.

### Actual Behavior
In the code when the ATR is calculated the values are backward filled with the future values and according to our previous example if a trade is opened in the period 10, then the ATR used is the ATR in the period 101 

```
def set_atr_periods(self, periods: int = 100):
    """
    Set the lookback period for computing ATR. The default value
    of 100 ensures a _stable_ ATR.
    """
    h, l, c_prev = self.data.High, self.data.Low, pd.Series(self.data.Close).shift(1)
    tr = np.max([h - l, (c_prev - h).abs(), (c_prev - l).abs()], axis=0)
    atr = pd.Series(tr).rolling(periods).mean().bfill().values  # Is look-ahead bias introduced using bfill?
    
    self.__atr = atr
```python

### Additional info

<!-- screenshots, code snippets, ... -->
<img width="302" alt="image" src="https://user-images.githubusercontent.com/39393035/232176952-fec49c22-01aa-4973-90f3-108c5402bab7.png">

It maintins the ATR calculation and changes until the period 101

My code is the following

class MyStrategy(TrailingStrategy):
    size = 0.3
    atr = 3
    atr_period = 100

    def init(self):
        super().init()
        self.set_trailing_sl(n_atr=self.atr)
        self.set_atr_periods(periods=self.atr_period)

    def next(self):
        super().next()
        
        if not self.position.is_long:
            self.buy(size=self.size)

**Comment by bravegag** [comment]
A simple fix is to do a `ffill` before the `bfill` like this:
```
atr = pd.Series(tr).rolling(periods).mean().ffill().bfill().values 
```
this ensures that only the leading (oldest prices chronologically) are backfilled for the purpose of calculating HPR.

bravegag:
@EladioRocha thanks a lot for reporting this. Can you please double-check my PR https://github.com/kernc/backtesting.py/pull/1230 to make sure that it works for you now?

kernc:
> this ensures that only the leading (oldest prices chronologically) are backfilled for the purpose of calculating HPR.

[...]

PR #1230, opened 2025-02-26
Implementation for issue #963

bravegag:
The issue is fixed by doing a ffill prior to the bfill so that bfill will only be applied to leading NAs, this way makes it possible to compute holding period return (HPR) from p[0] if needed. The ffill and bfill were moved prior to the rolling step to make sure there are no NAs and therefore adding the min_period=1 is not necessary to the rolling function.


JamesonDiamond1 (approved):
@JamesonDiamond1

PR #1276, opened 2025-05-27
Include Commissions in _trades DataFrame.

abhirathmahipal:
Include `Commissions` in `_trades` DataFrame.

### DIFF
--- a/backtesting/_stats.py
+++ b/backtesting/_stats.py
@@ -67,6 +67,7 @@ def compute_stats(
             'SL': [t.sl for t in trades],
             'TP': [t.tp for t in trades],
             'PnL': [t.pl for t in trades],
+            'Commissions': [t._commissions for t in trades],
             'ReturnPct': [t.pl_pct for t in trades],
             'EntryTime': [t.entry_time for t in trades],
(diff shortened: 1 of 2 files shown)


backtesting/_stats.py
```diff
@@ -67,6 +67,7 @@ def compute_stats(
             'SL': [t.sl for t in trades],
             'TP': [t.tp for t in trades],
             'PnL': [t.pl for t in trades],
+            'Commissions': [t._commissions for t in trades],
```
kernc, line 70: Reasonable! Can we put this in conditionally, only if `commission=` was passed non-0 which is the default?

kernc: I realize this would make sense for 'SL' and 'TP' too ...

abhirathmahipal: Also, `PnL` doesn't account for `Commissions`. Rationale behind this decision?

abhirathmahipal: Will push these fixes :)

abhirathmahipal:
Incrementally taking up the changes. Please review [this CR which includes commissions](https://github.com/kernc/backtesting.py/pull/1277) as the first step.

PR #1278, opened 2025-05-28
Introduce NetPnL in Position and Trades.

abhirathmahipal:
It's a good idea to maintain `pl` and `net_pl` separately. Reasons being

- `commission` is directly deducted from cash once a trade is entered.
- Clients can use `pl` and `net_pl` as per their use case.

If you're on the same page, will make similar changes for `pl_pct`.


abhirathmahipal:
I was thinking along your lines earlier. 

Reasoning for the above approach

- `Trade._commission` is set only after the Trade is closed. If the Strategy were use `Trade.pl`, it wouldn't reflect the commission.

Have raised a new pull request based on your inputs.

PR #1279, opened 2025-05-28
Apply Commissions to PnL and PnL Percentage.

abhirathmahipal:


backtesting/backtesting.py
```diff
@@ -671,9 +671,12 @@ def is_short(self):
 
     @property
     def pl(self):
-        """Trade profit (positive) or loss (negative) in cash units."""
+        """
+            Trade profit (positive) or loss (negative) in cash units.
+            Commissions are reflected only after the Trade is closed.
```
kernc, line 676:
```suggestion
        Trade profit (positive) or loss (negative) in cash units.
        Commissions are reflected only after the Trade is closed.
```

abhirathmahipal:
> Why not apply the same subtraction (`-self._commissions`) to `Trade.pl_pct` though? I.e. to keep the two PL values in sync. 🟦

This would work for `Broker` but not for `Trade`.  If parity has to be maintained in `Trade`, we'll have to factor commission into per unit price.

backtesting/backtesting.py
abhirathmahipal: Done!

abhirathmahipal:
Let's merge this?

As mentioned above 
> Will take up pl_pct and commission calculation bugs in the next CR.

abhirathmahipal:
>  The implementation is convoluted and bug-prone, but it's been like that before. 

That's the reason I just wanted to go for `PnL`.

PR #1302, opened 2025-08-12
Introduce flexible height by proportially sizing elements

polyclick:
Let's user set a hardcoded value for the main OHLC chart, will proportially resize other sections.

- Defaults to 400px
- Tested & works in Jupyter Notebooks
- Not tested in a browser

```
bt.plot(
  plot_height=800
)
```

SCR-20250812-iiwj-2
SCR-20250812-iiwj-3

### DIFF
--- a/backtesting/_plotting.py
+++ b/backtesting/_plotting.py
@@ -190,7 +190,7 @@ def _maybe_resample_data(resample_rule, df, indicators, equity_data, trades):
 def plot(*, results: pd.Series,
          df: pd.DataFrame,
          indicators: List[_Indicator],
-         filename='', plot_width=None,
+         filename='', plot_width=None, plot_height=400,
          plot_equity=True, plot_return=False, plot_pl=True,
          plot_volume=True, plot_drawdown=False, plot_trades=True,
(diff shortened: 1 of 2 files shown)


backtesting/_plotting.py
```diff
@@ -207,6 +216,16 @@ def plot(*, results: pd.Series,
...
 
+    # Calculate proportional heights based on original ratios
+    # Original OHLC height was 400px, use that as base ratio
+    base_height = plot_height
+    equity_height = int(base_height * 0.25)  # was 100px
+    equity_height_with_dd = int(base_height * 0.2)  # was 80px
+    drawdown_height = int(base_height * 0.2)  # was 80px
+    pl_height = int(base_height * 0.2)  # was 80px
+    volume_height = int(base_height * 0.175)  # was 70px
+    indicator_height = int(base_height * 0.125)  # was 50px
```
kernc, line 227: I don't like that these are fixed proportions, and that's that.

Maybe we could make param `plot_height: int | Tuple[int] | Dict[str, int]`? :thinking:

kernc (changes requested):

PR #1316, opened 2025-10-05
[Doc] clarity on trade pct calculation

### LINKED ISSUE #1254: More clarity on trade pct calculation
Hello. 

The `stats._trades` has a column called `ReturnPct`. By reading the docs and the source code it's not clear if the return in percentage is calculated relative to the starting capital, the current equity, or the trade volume. By reading the code:
```python
    @property
    def pl(self):
        """Trade profit (positive) or loss (negative) in cash units."""
        price = self.__exit_price or self.__broker.last_price
        return self.__size * (price - self.__entry_price)

    @property
    def pl_pct(self):
        """Trade profit (positive) or loss (negative) in percent."""
        price = self.__exit_price or self.__broker.last_price
        return copysign(1, self.__size) * (price / self.__entry_price - 1)

```

it's clear that the return Pct comes from the trade volume. So it would be useful to let the user know this in the docs:
```python
"""Trade profit (positive) or loss (negative) in percent relative to trade size.""" 
```

Thank's :)

ved1beta:
closes #1254

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -681,7 +681,7 @@ class Trade:
 
     @property
     def pl_pct(self):
-        """Trade profit (positive) or loss (negative) in percent."""
+        """Trade profit (positive) or loss (negative) in percent relative to trade entry price."""
         price = self.__exit_price or self.__broker.last_price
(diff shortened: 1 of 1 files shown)


kernc:
Thanks! If this is for hacktoberfest, I hope you made it! 😆

Issue #1318, opened 2025-10-05
CGPT5 proposed enhancements


bravegag:
### Enhancement description

Yes, a few likely issues stand out:

- Strategy parameter clobbering: Strategy.check_params sets attributes directly. A user could pass a parameter starting with “” (e.g., broker) and overwrite internals. Block names starting with “” or keep a denylist.
- Potential infinite reprocessing: _Broker._process_orders can recursively call itself when reprocess_orders is True. In pathological cases this could loop. Add a small recursion/depth guard.
- Ambiguous price in validation errors: _Broker.new_order validates SL/LIMIT/TP against an “adjusted” price but error messages don’t show which price was used. This is confusing. Include the reference price used (limit/stop/adjusted market).
- Weak diagnostics on canceled orders: When canceling due to insufficient margin (both proportional size resolving to zero and absolute size exceeding margin), there’s no contextual info. Emitting warnings with size, price, margin_available, and leverage would aid debugging.
- Indicator overlay heuristic: In Strategy.I, overlay autodetection does value/Close broadcasting for 2D indicators; while it generally works, odd dtypes or shapes can mislead. Consider stricter numeric checks or clearer error when shapes don’t align._

### Code sample

```python

```

OllyNewport:
Hi! I'd be interested in contributing to this issue as a first contribution. Are all of the proposed enhancements still wanted?

kernc:
Any that are still relevant, sure. Best to just let your agent decide. 😈

OllyNewport:
Hi @kernc — just wanted to check in on this whenever you get a chance to look, along with the other two I've got open on the repo (#1391, #1392). Happy to make any changes if something needs adjusting. Thanks for maintaining the project!

PR #1319, opened 2025-10-05
Implementation for issue #1318

bravegag:
The patch corresponds to issue #1318, auto-AI generated and it does change the backtesting result. It should be reviewed in detail and some changes may be valid others perhaps not.

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -16,6 +16,7 @@ from functools import lru_cache, partial
 from itertools import chain, product, repeat
 from math import copysign
 from numbers import Number
+from difflib import get_close_matches
 from typing import Callable, List, Optional, Sequence, Tuple, Type, Union
 
 import numpy as np
@@ -64,10 +65,12 @@ class Strategy(metaclass=ABCMeta):
     def _check_params(self, params):
         for k, v in params.items():
             if not hasattr(self, k):
+                suggestions = get_close_matches(k, [a for a in dir(self) if not a.startswith('_')], n=3)
+                hint = f" Did you mean: {', '.join(suggestions)}?" if suggestions else ""
                 raise AttributeError(
-                    f"Strategy '{self.__class__.__name__}' is missing parameter '{k}'."
+                    f"Strategy '{self.__class__.__name__}' is missing parameter '{k}'. "
                     "Strategy class should define parameters as class variables before they "
-                    "can be optimized or run with.")
+                    "can be optimized or run with." + hint)
             setattr(self, k, v)
         return params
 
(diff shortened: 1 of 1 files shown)


backtesting/backtesting.py
```diff
@@ -139,6 +142,8 @@ def _format_name(name: str) -> str:
 
         try:
             value = func(*args, **kwargs)
+            if isinstance(value, pd.Series):
+                value = value.to_numpy()
```
kernc, line 146: This case is covered a few lines further below ... 🤔
https://github.com/kernc/backtesting.py/blob/cc1ebc8953fbeeb1904345171e01fd804ce071c2/backtesting/backtesting.py#L153-L154

backtesting/backtesting.py
```diff
@@ -448,7 +453,7 @@ def __repr__(self):
                                                  ('tp', self.__tp_price),
                                                  ('contingent', self.is_contingent),
                                                  ('tag', self.__tag),
-                                             ) if value is not None))  # noqa: E126
+                                             ) if (value is not None and (not isinstance(value, bool) or value))))  # noqa: E126
```
kernc, line 456: Covered by the use of `try_`, fixed in 8dd1e369.

backtesting/backtesting.py
```diff
@@ -998,6 +1005,10 @@ def _process_orders(self):
             # If we don't have enough liquidity to cover for the order, the broker CANCELS it
             if abs(need_size) * adjusted_price_plus_commission > \
                     self.margin_available * self._leverage:
+                warnings.warn(
+                    f'({data.index[self._i]}) broker canceled order {order} due to insufficient margin '
+                    f'(equity={self.equity:.2f}, margin_available={self.margin_available:.2f}).',
+                    category=UserWarning)
```
kernc, line 1011: This hunk is legit, I think. :+1: If it can be made to match the prior art in https://github.com/kernc/backtesting.py/commit/5e72e7b270a468ae3747c36678292f6469f48f82 ...

kernc (changes requested):
In general, I'm finding most people's LLMs presently are doing lousy jobs with full projects' sources. But fascinatingly, while the PR contains no items listed in https://github.com/kernc/backtesting.py/issues/1318 (🤔), there are a few bits here we would like to keep! 😃

bravegag:
> In general, I'm finding most people's LLMs presently are doing lousy jobs with full projects' sources. But fascinatingly, while the PR contains no items listed in #1318 (🤔), there are a few bits here we would like to keep! 😃

Indeed, CGPT went wild on it and changed the functional behavior of the backtest. I wanted to keep it simple and avoid going down a rabbit hole of changes so trimmed it to the minimum valid changes without breaking the current backtest behavior.

PR #1336, opened 2025-12-16
Feature/upgrade uv

iyedexe:
Upgraded python version to 3.10 and more
Using ruff for formatting
Using pyright for type checking
Using uv for dependencies and package build
typing using native dict instead of Dict and all other types.


kernc:
To its considerable advantage, this project does not use uv.

PR #1346, opened 2026-02-12
BUG: Fix CAGR and annualized return calculations

truthyblue:
## Summary

- **CAGR**: `time_in_years` was computed by dividing calendar days by `annual_trading_days` (e.g. 252, 52, 12) instead of 365.25. This inflated the time period and underreported CAGR. For weekly/monthly data, results were drastically wrong (e.g. 365/52 = 7 years instead of ~1 year).
- **Return (Ann.)**: `pct_change()` produces a leading NaN that `geometric_mean()` counted in the denominator via `fillna(0)`, slightly diluting the geometric mean and underestimating annualized return.

After the fix, CAGR and Return (Ann.) are now consistent with each other as expected (21.16% vs 21.19% on GOOG SmaCross, previously 14.16% vs 21.18%).

## Test plan

- [x] Updated expected values in `test_compute_stats`
- [x] All existing tests pass (74 passed, 1 skipped; `test_FractionalBacktest` is a pre-existing failure unrelated to this change)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


TheNoobiCat:
@copilot is this a real issue or just made up ai slop what do you think bro

atharvajoshi01 (commented):
Two solid fixes here.

The extra `.dropna()` after `pct_change()` makes sense since the first return is always NaN and would mess with `geometric_mean`.

Using 365.25 for CAGR is the correct convention - CAGR should use calendar days, not trading days. The old code was dividing by `annual_trading_days` (252) which overstates the compounding period and underreports the actual CAGR. The jump from 14.15% to 21.16% in the test reflects this correction.

kernc:
Lgtm, thanks!

PR #1358, opened 2026-04-14
Fix MultiBacktest when some runs make no trades

### LINKED ISSUE #1344: MultiBacktest.run errors out when tested strategy sometimes never enters a position
### Expected behavior

MultiBacktest.run should correctly handle when strategy chooses not to enter any positions and return a sensible dataframe

### Code sample

```python
import random

# will_buys = [True, True, True, True, True] # works, returns series
# will_buys = [False, False, False, False, False] # works, returns series of None
will_buys = [True, False, False, False, False] # errors
# will_buys = [False, True, False, False, False] # works but returns a weird nested series thing
# note the effect of (not) buying in the first tested instrument
class TestStrat(Strategy):
    def init(self):
        self.will_buy = will_buys[len(self.data.index) - 2144]
        self.has_bought = False

    def next(self):
        if not self.will_buy:
            return
        
        if self.position:
            self.position.close()
        
        if not self.has_bought:
            self.buy()
            self.has_bought = True
        

print(will_buys)

bt = MultiBacktest([GOOG[:-4],GOOG[:-3],GOOG[:-2],GOOG[:-1],GOOG], TestStrat,
               cash=10000, commission=.002,
               exclusive_orders=True)
bt.run()
```

### Actual behavior

MultiBacktest.run errors out with `TypeError: object of type 'NoneType' has no len()` or returns an undocumented nested dataframe(?) object

_No response_

### Software versions

- `backtesting.__version__`: 0.6.5
- `pandas.__version__`: 3.0.0
- `numpy.__version__`: 2.4.1
- `bokeh.__version__`: 3.8.2
- OS: Windows 11

BerqiaMouad:
## Summary
- keep the existing public stats series for zero-trade runs instead of replacing it with `None`
- add regression coverage for all-zero and mixed zero/non-zero `MultiBacktest` runs

## Why
`Backtest.run()` already returns meaningful public stats when a strategy makes zero trades.
Dropping those results to `None` makes `MultiBacktest.run()` either raise when a batch mixes `Series` and `None`, or return a one-row frame of `None` values.

Fixes #1344.
Closes https://github.com/kernc/backtesting.py/pull/1366

## Validation
- `python -m unittest -v backtesting.test._test.TestLib.test_MultiBacktest backtesting.test._test.TestLib.test_MultiBacktest_keeps_zero_trade_runs`
- `flake8 backtesting setup.py`

## Notes
- On my local macOS / Python 3.11.14 / pandas 3.0.2 setup, `python -m backtesting.test` still hits the pre-existing `test_FractionalBacktest` failure (`ValueError: output array is read-only`).

### DIFF
--- a/backtesting/lib.py
+++ b/backtesting/lib.py
@@ -608,7 +608,7 @@ class MultiBacktest:
         data_shm, strategy, bt_kwargs, run_kwargs = args
         dfs, shms = zip(*(SharedMemoryManager.shm2df(i) for i in data_shm))
         try:
-            return [stats.filter(regex='^[^_]') if stats['# Trades'] else None
+            return [stats.filter(regex='^[^_]')
                     for stats in (Backtest(df, strategy, **bt_kwargs).run(**run_kwargs)
                                   for df in dfs)]
         finally:
--- a/backtesting/test/_test.py
+++ b/backtesting/test/_test.py
@@ -4,6 +4,7 @@ import os
 import sys
 import time
 import unittest
+import warnings
 from concurrent.futures.process import ProcessPoolExecutor
 from contextlib import contextmanager
 from glob import glob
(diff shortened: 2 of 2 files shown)


atharvajoshi01:
Fix seems good idea, as there's no reason to discard zero trade stats _`backtest.run()`_ already returns valid data for those. Tests looks go tbh.

kernc:
Even as this PR was proposed a bit sooner, I liked the other simpler test better. Many thanks to everyone involved for figuring it out!

PR #1364, opened 2026-06-16
Clarify commission parameter in README usage example

mastroke:
## Summary
- Adds a short note below the README usage snippet explaining `commission=.002` as a 0.2% per-trade fee
- Points readers to `spread` and callable commission in the API reference

## Motivation
New users often miss that the commission argument is a fraction of traded value. The quick-start docs mention 0.2% in prose, but the README code sample does not.

## Test plan
- [x] Docs-only change

Made with [Cursor](https://cursor.com)

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,10 @@ stats = bt.run()
 bt.plot()
 ```
 
+In this example, `commission=.002` applies a **0.2% fee per trade**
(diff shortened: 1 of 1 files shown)


kernc:
The readme is meant to be brief, glance-friendly. Cursor may disagree, but the interested user is expected to RTFM. Other parameters are not similarly explained, etc.

PR #1368, opened 2026-07-10
Fix trailing ATR look-ahead bias

### LINKED ISSUE #963: The atr introduce look ahead bias or is my interpretation incorrect?

Jaco-Ren:
## Summary

- stop backfilling `TrailingStrategy` ATR values, which used future volatility during the rolling warmup window
- skip trailing stop updates until ATR is finite instead of setting a stop from unavailable data
- keep `set_trailing_pct()` stable by averaging only finite close/ATR ratios
- add a regression test proving a trade opened before ATR warmup does not receive a future-derived stop loss

## Motivation and Context

Fixes #963.

`TrailingStrategy.set_atr_periods()` previously used `.bfill()` after the rolling ATR calculation. That filled the warmup bars with ATR values computed from later bars, so strategies could open early trades whose trailing stop used future information.

With this change, early ATR entries remain `NaN`. `TrailingStrategy.next()` ignores those bars for stop updates and starts applying trailing stops once the rolling ATR window is actually available.

## Testing

- `python -m pytest backtesting/test/_test.py -q`
- `git diff --check`
- `python -m py_compile backtesting\lib.py backtesting\test\_test.py`

### DIFF
--- a/backtesting/lib.py
+++ b/backtesting/lib.py
@@ -470,7 +470,7 @@ class TrailingStrategy(Strategy):
         """
         hi, lo, c_prev = self.data.High, self.data.Low, pd.Series(self.data.Close).shift(1)
         tr = np.max([hi - lo, (c_prev - hi).abs(), (c_prev - lo).abs()], axis=0)
-        atr = pd.Series(tr).rolling(periods).mean().bfill().values
+        atr = pd.Series(tr).rolling(periods).mean().values
         self.__atr = atr
 
     def set_trailing_sl(self, n_atr: float = 6):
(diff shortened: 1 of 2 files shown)


backtesting/lib.py
```diff
@@ -490,20 +490,25 @@ def set_trailing_pct(self, pct: float = .05):
             with `mean(Close * pct / atr)` and set with `set_trailing_sl`.
         """
         assert 0 < pct < 1, 'Need pct= as rate, i.e. 5% == 0.05'
-        pct_in_atr = np.mean(self.data.Close * pct / self.__atr)  # type: ignore
+        ratios = self.data.Close * pct / self.__atr  # type: ignore
+        ratios = ratios[np.isfinite(ratios)]
+        pct_in_atr = np.mean(ratios) if len(ratios) else np.nan
```
kernc, line 495: The existing line could be wrapped simpler?
```py
with np.errstate(invalid='ignore'):
    ...
```

backtesting/lib.py
```diff
@@ -490,20 +490,25 @@ def set_trailing_pct(self, pct: float = .05):
...
         self.set_trailing_sl(pct_in_atr)
 
     def next(self):
         super().next()
         # Can't use index=-1 because self.__atr is not an Indicator type
         index = len(self.data) - 1
+        atr = self.__atr[index]
+        if not np.isfinite(atr):
```
kernc, line 503:
```suggestion
        if np.isnan(atr):
```

kernc (changes requested):

Issue #1288, opened 2025-07-19
Stop Loss value being erased when reached


xyffar:
### Expected behavior

The SLs should be displayed in the trade list (within the result stats) whenever they are set (according to the strategy).

<img width="349" height="167" alt="Image" src="https://github.com/user-attachments/assets/cd4a35f7-e87d-4025-bec3-d9f361828a64" />

### Actual behavior

When I get the list of the trades, some of them are unprovided with the SL. This seems to happen for the trades being closed for reaching the SL.
<img width="502" height="167" alt="Image" src="https://github.com/user-attachments/assets/73dcc223-c352-4a67-8c3d-22b8811263d6" />

### Additional info, steps to reproduce, full crash traceback, screenshots

I analyzed the source code and the SL values are correctly passed within the `strategy.buy()` method. I followed the code and the values are actually erased when `order._replace(stop_price=None)` is called (due to the the stop loss being reached).
https://github.com/kernc/backtesting.py/blob/f8f49694472b300168bf3dadee341fcd38e09188/backtesting/backtesting.py#L883-L892

The reason is the `stop_price` parameter , aka `order.stop`, having the same `id `of the `order.parent_trade.sl`. The "coupling" arises within the `__set_contingent` method, line 738-739, where we create the stop order by passing the `price` variable (and its reference), which is the SL of the order. In brief, the contingent stop order has the attribute `stop` with the same reference of the attribute `sl` of the parent order. When the former gets None'd, the latter follows it.

https://github.com/kernc/backtesting.py/blob/f8f49694472b300168bf3dadee341fcd38e09188/backtesting/backtesting.py#L730-L740

As proof, see below where we can see that the two variables point to the same address.
<img width="1114" height="345" alt="Image" src="https://github.com/user-attachments/assets/8730c494-0c1a-402a-8766-6093d9577158" />

### Software versions

- backtesting 0.6.4
- pandas 2.2.3
- numpy 1.26.0
- bokeh 3.7.3

PR #1369, opened 2026-07-11
BUG: Preserve stop-loss value in stats._trades when SL is gapped through

### LINKED ISSUE #1340: Stop Loss value being (still sometimes) erased when reached
### Expected behavior

The SLs should be displayed in the trade list (within the result stats) whenever they are set (according to the strategy).
Image

### Code sample

```python
from backtesting.lib import crossover

class Strat(Strategy):
    def init(self):
        pass
    def next(self):
        if self.position:
            pass
        else:
            price = self.data.Close[-1]
            self.buy(size = 1, sl = price - 10, tp = price + 20)
        
bt = Backtest(GOOG, Strat, cash=10_000)
output = bt.run()
display(output._trades.iloc[30:33])
bt.plot()
```

### Actual behavior

When I get the list of the trades, some of them are unprovided with the SL. This seems to happen **only** for the trades being closed **BELOW** the set SL.

This issue seems to be an unfixed part of the bug report https://github.com/kernc/backtesting.py/issues/1288 by @xyffar. 

Image

### Software versions

- `backtesting.__version__`: 0.6.5
- `pandas.__version__`: 2.3.3
- `numpy.__version__`: 1.26.2
- `bokeh.__version__`: 3.8.1
- OS: Linux 6.14.0-36-generic x86_64

### LINKED ISSUE #1370: Order executed with an unreachable price.
### Expected behavior

Order should executed when price reaches 160 some day else.

Image

### Code sample

```python
import pandas as pd

class MyStrategy(Strategy):
    def init(self):
        pass

    def next(self):
        if not self.position:
            if self.data.datetime == pd.Timestamp('2004/10/21'):
                self.buy(stop=160)


GOOG['datetime'] = GOOG.index

bt = Backtest(GOOG, MyStrategy,
              cash=100000, commission=0.003,
              exclusive_orders=True,
              trade_on_close=True,
              finalize_trades=True
              )

output = bt.run()
print(output['_trades'])
bt.plot()
```

### Actual behavior

Order executed with price 160 next day which wont happen in practice.

print(output['_trades']) prints
   Size  EntryBar  ExitBar  EntryPrice  ExitPrice    SL    TP       PnL  ReturnPct  EntryTime   ExitTime  Duration   Tag
0   623        45     2146       160.0      801.2  None  None  399467.6     4.0075 2004-10-22 2013-02-28 3051 days  None

_No response_

### Software versions

- `backtesting.__version__`: '0.6.5'
- `pandas.__version__`: '2.2.3'
- `numpy.__version__`: '2.2.3'
- `bokeh.__version__`: '3.7.0'
- OS:

nyxst4ck:
### Problem

When a trade is closed by its stop-loss, the SL value is sometimes missing (`NaN`) in `stats._trades["SL"]`, even though an SL was explicitly set. This happens specifically when the price **gaps through** the stop: the bar opens beyond the stop price, so the fill price is worse than the stop.

This is the still-unfixed part of #1288 reported in #1340.

### Root cause

When a stop order is hit, the stop becomes a market order and its `stop_price` is cleared:

```python
order._replace(stop_price=None)
```

Since the trade reads its SL from `trade._sl_order.stop`, clearing it drops the value from the trades data frame. The previous fix (#1288) restored it, but only when the fill price equalled the stop exactly:

```python
if price == stop_price:
    trade._sl_order._replace(stop_price=stop_price)
```

On a gap-through, the order fills at the (worse) open price, so `price != stop_price` and the SL is never restored.

### Fix

Restore the recorded SL whenever the executed order is the trade's own stop-loss order, regardless of the fill price. The `TP`/`trade.close()` paths are unaffected (those orders carry no stop price).

### Test

Added `test_sl_value_in_trades_df_when_gapped_through`, which enters a long position whose SL (99.5) is gapped through on the next bar's open (99.19). The exit price is below the stop, but the SL value must still be recorded.

- Red on `master`: `AssertionError: None != 99.5`
- Green with this change.

Existing SL/TP regression tests (`test_sl_tp_values_in_trades_df`, `test_sl_always_before_tp`, `test_stop_entry_and_tp_in_same_bar`) still pass.

Fixes #1340

Fixes https://github.com/kernc/backtesting.py/issues/1370

### DIFF
--- a/backtesting/test/_test.py
+++ b/backtesting/test/_test.py
@@ -1254,8 +1254,8 @@ class TestRegressions(TestCase):
         trades = Backtest(SHORT_DATA, S).run()._trades
         self.assertEqual(len(trades), 1)
         trade = trades.iloc[0]
-        # The long SL (99.5) is gapped through on the next bar's open (99.19),
-        # so the exit price is below the stop price ...
-        self.assertLess(trade['ExitPrice'], 99.5)
+        # The long SL (99.5) is gapped through on the next bar's open, so the
+        # trade exits at that worse market price rather than at the stop price.
+        self.assertEqual(trade['ExitPrice'], 99.19)
         # ... yet the SL value must be preserved in the trades data frame.
         self.assertEqual(trade['SL'], 99.5)
--- a/backtesting/test/_test.py
+++ b/backtesting/test/_test.py
@@ -1177,3 +1177,21 @@ class TestRegressions(TestCase):
         trades = Backtest(SHORT_DATA, S).run()._trades
         self.assertEqual(trades['SL'].fillna(0).tolist(), [0, 99])
         self.assertEqual(trades['TP'].fillna(0).tolist(), [111, 0])
+
+    def test_sl_value_in_trades_df_when_gapped_through(self):
+        # An SL that is gapped through (the bar opens beyond the stop, so the
+        # fill price is worse than the stop price) must still be recorded in
+        # stats._trades["SL"]. See GH issue #1340.
+        class S(_S):
+            def next(self):
+                if len(self.data.index) == 9:
+                    self.buy(size=1, sl=99.5)
+
+        trades = Backtest(SHORT_DATA, S).run()._trades
+        self.assertEqual(len(trades), 1)
+        trade = trades.iloc[0]
+        # The long SL (99.5) is gapped through on the next bar's open (99.19),
+        # so the exit price is below the stop price ...
+        self.assertLess(trade['ExitPrice'], 99.5)
+        # ... yet the SL value must be preserved in the trades data frame.
+        self.assertEqual(trade['SL'], 99.5)
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -916,8 +916,11 @@ class _Broker:
                 if trade in self.trades:
                     self._reduce_trade(trade, price, size, time_index)
                     assert order.size != -_prev_size or trade not in self.trades
-                    if price == stop_price:
+                    if order is trade._sl_order:
                         # Set SL back on the order for stats._trades["SL"]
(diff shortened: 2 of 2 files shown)


nyxst4ck:
Marked ready for review. I do not see any checks reported on this branch yet.

backtesting/test/_test.py
```diff
@@ -1241,3 +1241,21 @@ def next(self):
...
+            def next(self):
+                if len(self.data.index) == 9:
+                    self.buy(size=1, sl=99.5)
+
+        trades = Backtest(SHORT_DATA, S).run()._trades
+        self.assertEqual(len(trades), 1)
+        trade = trades.iloc[0]
+        # The long SL (99.5) is gapped through on the next bar's open (99.19),
+        # so the exit price is below the stop price ...
+        self.assertLess(trade['ExitPrice'], 99.5)
```
kernc: Shouldn't the exit price still be 99.19? You don't get a better offer than what the market opens with!

Similar issue: https://github.com/kernc/backtesting.py/pull/1373#issuecomment-5024487173

nyxst4ck: Yes — the trade still fills at the 99.19 market open. This PR only preserves the original 99.5 stop in the SL stats field after that fill. I tightened the regression to assert ExitPrice == 99.19 and SL == 99.5; the focused test passes, and the full 78-test suite completed successfully. Pushed in 79fd3f1.

kernc: Gah, I misread the original assertion: it's `assertLess`, so it was fine. 👌 A LLM wouldn't make such a trivial mistake.

It's now clearer, though.

kernc:
Many thanks! If you find any other similarly neat fixes, they are welcome!

PR #1373, opened 2026-07-20
BUG: Fill gapped stop orders at the open

### LINKED ISSUE #1370: Order executed with an unreachable price.

Sanjays2402:
Fixes #1370

With `trade_on_close=True`, triggered stop orders used the previous close as their fill-price basis, allowing a buy stop to fill below a gapped-up open. Stop orders now use the next bar's open while plain market orders retain trade-on-close behavior; the regression covers a 105 stop followed by a 110 open.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 on:
   push: { branches: [master] }
   pull_request: { branches: [master] }
-  schedule: [ cron: '2 2 * * 6' ]  # Every Saturday, 02:02
+  schedule: [ cron: '2 2 1 * *' ]
 
 env:
   TQDM_MININTERVAL: 10
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 1
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd  # v5
       - run: pip install flake8 mypy
       - run: flake8 backtesting setup.py
       - run: mypy --no-warn-unused-ignores backtesting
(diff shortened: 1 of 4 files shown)


kernc:
I'm not convinced the proposed fix is correct.

Most brokers work like this: You set a buy-stop at ¤105. The market session opens at ¤110. Your order gets filled at **¤110**. This is the [gap risk](https://www.investopedia.com/terms/g/gaprisk.asp).

Sanjays2402:
You're right, and thanks for the gap-risk link — I had the fill semantics backwards. A stop that gaps through should fill at the open, not at the stop price, so my change models something brokers don't actually do. The "bug" I was chasing is really correct gap behavior.

I'll withdraw this rather than leave a wrong fix sitting open. If there's a genuine edge case hiding in here (e.g. the fill price when the open is on the *other* side of the stop), I'll reopen with a proper repro. Appreciate you taking the time to walk through it.

PR #1374, opened 2026-07-20
CI: Set PYTHONWARNINGS=always::FutureWarning,always::DeprecationWarning

kernc:


kernc:
@copilot Investigate why this PR fails.

kernc:
@copilot The diagnosis was correct, but the proposed fix not so much. We want to error on all stray warnings and likewise want to have all warnings accounted for.
Can you fix the remaining test failure in `test_FractionalBacktest` which seemingly won't take simple `self.assertWarns()`.

kernc:
@copilot Again, your insight was valuable and correct, leading to a neat bugfix, but your patch was inadequate. Anyway, thanks. The case is closed with success.

-----
Hope this change doesn't bite us later on when the platform (CI runners) or the upstreams begin emitting warnings non-deterministically. But at least we'll learn about it early.

PR #1375, opened 2026-07-20
BRK: Apply spread at trade entry and exit

### LINKED ISSUE #1335: Apply half-spread and commission on both open/close trade
### Enhancement description

Hi,

Full spread is currently applied to buy price on trade open and open+close commission only on trade close. This doesn't reflect real life and affects backtesting stop loss/trailing stop loss trigger level.
Also,, I'm using trade.pl_pct for trailing stop loss calculation and I have to subtract commission manually.
It also leads to bad optimization. 

The situation should be :
- trade open: apply half-spread on buy price and also apply commission
- trade close: apply half-spread on sell price and also apply commission

Thanks a lot!

### Code sample

```python

```

### Additional info, images

_No response_

nyxst4ck:
Closes #1335.

## Summary

- apply the configured bid-ask spread adversely at both trade entry and trade exit
- retain the existing commission model, including recalculation for each transaction
- document the two-sided spread behavior
- cover long/short round trips, reversals, stop-loss fills, and updated equity expectations

This revision follows the maintainer feedback: the spread is no longer split in half, and the commission accounting changes from the earlier draft have been removed.

## Validation

- `python -m backtesting.test` — 81 passed, 1 skipped
- `git diff --check`

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -349,7 +349,7 @@ class Position:
     @property
     def pl(self) -> float:
         """Profit (positive) or loss (negative) of the current position in cash units."""
-        return self.__broker._position_unrealized_pl
+        return sum(trade.pl for trade in self.__broker.trades)
 
     @property
     def pl_pct(self) -> float:
@@ -656,7 +656,7 @@ class Trade:
     def pl(self):
         """
         Trade profit (positive) or loss (negative) in cash units.
-        Commissions are reflected only after the Trade is closed.
+        Commissions already incurred are reflected.
         """
         price = self.__exit_price or self.__broker.last_price
         return (self.__size * (price - self.__entry_price)) - self._commissions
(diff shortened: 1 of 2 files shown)


kernc:
@nyxst4ck Thanks for picking this up!

> 1. split the configured bid-ask spread equally between trade entry and exit

Modelling spread as a share of price is somewhat eccentric and variable already. But then "you need to further compute halves" will be a documentation nightmare, leaving the logic all but simple. Spread is property of the instrument and it affects us, in whole, when entering the trade.

Then it affects us again when we try to sell the gold! Thus, to be precise, I'm thinking spread should also be accounted for twice. 🤔😳

**But this shall break a lot of existing expected values.**

But breaking spread in half is a decisive no-go. :-1:  

> 3. preserve the commission actually charged at entry instead of recomputing it at close

This is also not how exchanges do. The commission, for what I see, is computed again based on the order value at the time of the transaction. 

> 4. allocate entry commission proportionally when a trade is partially closed

Fixing 3. proper makes this a no-op.

nyxst4ck:
Makes sense, thanks for the detailed explanation. Reworked: spread is now applied in full at both entry and exit (`_adjusted_price` no longer halves it), and the proportional entry-commission allocation on partial closes is gone — commission is simply recomputed per transaction, same as before. `python -m backtesting.test`: 81 passed, 1 skipped.

PR #1376, opened 2026-07-21
DOC: Clarify fractional order size semantics

initial-d:
Clarifies the `Order.size` documentation for fractional sizes.

The existing docs correctly say that `0 < size < 1` is interpreted as a fraction of current available liquidity. This small addition makes explicit that fractional sizing controls order cash allocation, not the maximum account risk if a stop-loss is hit.

This is a documentation-only change. I ran:

```bash
python -m py_compile backtesting/backtesting.py
```

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -454,6 +454,8 @@ class Order:
         If size is a value between 0 and 1, it is interpreted as a fraction of current
         available liquidity (cash plus `Position.pl` minus used margin).
         A value greater than or equal to 1 indicates an absolute number of units.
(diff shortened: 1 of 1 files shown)


kernc:
I chose to reword the docstring a little bit. Please LMK if it's still clear enough.

initial-d:
Yes, this is still clear to me. The added "to allocate to the order" wording keeps the existing liquidity-sizing definition, and the parenthetical makes the risk-vs-allocation distinction explicit without overexplaining it. Thanks for polishing the wording.

PR #1377, opened 2026-07-22
TEST: Cover absolute-size margin warning

### LINKED ISSUE #1360: Silent order cancellation for absolute-sized orders with insufficient margin (no warning emitted)
### Expected behavior

A UserWarning should be emitted (matching the fractional-size path), e.g.:

UserWarning: time=N: Broker canceled the order (size=4) due to insufficient margin.

### Code sample

```python

```

### Actual behavior

Silent order cancellation

Description

When an order with absolute size (integer or float ≥ 1) is cancelled due to insufficient margin, the broker silently removes it with no feedback to the user. This is inconsistent with the behavior for fractional sizes (-1 < size < 1), which correctly emit a UserWarning.

Affected code (_Broker._process_orders):

# If we don't have enough liquidity to cover for the order, the broker CANCELS it
if abs(need_size) * adjusted_price_plus_commission > self.margin_available * self._leverage:
    self.orders.remove(order)
    continue  # no warning here
Compare with the fractional-size path, which does warn:

if not size:
    warnings.warn(
        f'time={self._i}: Broker canceled the relative-sized '
        f'order due to insufficient margin.', category=UserWarning)
    self.orders.remove(order)
    continue
How to reproduce

Run a backtest where the absolute position size × price exceeds available cash. For example, cash=100_000 with buy(size=4) at a price of ~27,000 (NQ futures-like data). The order is silently dropped every bar — no exception, no warning, no indication in stats.

Expected behavior

A UserWarning should be emitted (matching the fractional-size path), e.g.:

UserWarning: time=N: Broker canceled the order (size=4) due to insufficient margin.
Suggested fix

if abs(need_size) * adjusted_price_plus_commission > self.margin_available * self._leverage:
    warnings.warn(
        f'time={self._i}: Broker canceled the order (size={need_size}) '
        f'due to insufficient margin.', category=UserWarning)
    self.orders.remove(order)
    continue

### Software versions

- `backtesting.__version__`:
- `pandas.__version__`:
- `numpy.__version__`:
- `bokeh.__version__`:
- OS:

initial-d:
Adds a regression test for absolute-sized orders that are canceled because there is not enough margin.

This covers the behavior discussed in #1360. The implementation on current 'master' already emits a UserWarning for the absolute-size path; this test makes that behavior explicit and keeps it aligned with the existing relative-size warning path.

Validation:
- 'python -m py_compile backtesting/test/_test.py' passes
- 'git diff --check' passes
- Tried to run the targeted unittest locally, but the available Python environment was missing test/runtime dependencies such as 'bokeh'. I started a local .venv install, but dependency installation took too long in this environment, so I left full execution to CI.

Fixes https://github.com/kernc/backtesting.py/issues/1360

### DIFF
--- a/backtesting/test/_test.py
+++ b/backtesting/test/_test.py
@@ -231,6 +231,19 @@ class TestBacktest(TestCase):
                       cash=1000, spread=.01, margin=.1, trade_on_close=True)
         bt.run()
 
+    def test_absolute_size_order_warns_on_insufficient_margin(self):
+        class S(Strategy):
+            def init(self):
+                pass
+
+            def next(self):
+                self.buy(size=10_000)
+
+        bt = Backtest(GOOG.iloc[:3], S, cash=1000)
+        with self.assertWarnsRegex(UserWarning, 'insufficient margin'):
+            stats = bt.run()
+        self.assertEqual(stats['# Trades'], 0)
+
     def test_spread_commission(self):
         class S(Strategy):
             def init(self):


kernc:
Ah, thanks for checking (and proving) that!

initial-d:
Thanks for merging it, and thanks for the quick review.

PR #1385, opened 2026-08-05
BUG: Size shared memory by materialized array, not lazy Index.nbytes

### LINKED ISSUE #1237: bt.optimize() results in "buffer is too small for requested array"
### Expected behavior

When I try to optimize the parameters of the simple SMA cross example provided in the tutorials, but instead use custom 1min OHLC data for a period of one month (44640 units), the following code results in an error: `TypeError: buffer is too small for requested array`.
 
```python
stats = bt.optimize(
    n1=[10], n2=[20], maximize="Equity Final [$]", constraint=lambda param: param.n1 < param.n2
)
bt.plot(plot_equity=False, plot_return=True)
print(stats)
```

```python
TypeError                                 Traceback (most recent call last)
Cell In[19], line 1
----> 1 stats = bt.optimize(
      2     n1=[10], n2=[20], maximize="Equity Final [$]", constraint=lambda param: param.n1 < param.n2
      3 )
      4 bt.plot(plot_equity=False, plot_return=True)
      5 print(stats)

File ~/Coding/algotrading/.venv/lib/python3.12/site-packages/backtesting/backtesting.py:1630, in Backtest.optimize(self, maximize, method, max_tries, constraint, return_heatmap, return_optimization, random_state, **kwargs)
   1627     return stats if len(output) == 1 else tuple(output)
   1629 if method == 'grid':
-> 1630     output = _optimize_grid()
   1631 elif method in ('sambo', 'skopt'):
   1632     output = _optimize_sambo()

File ~/Coding/algotrading/.venv/lib/python3.12/site-packages/backtesting/backtesting.py:1527, in Backtest.optimize.<locals>._optimize_grid()
   1524     shm_refs.append(shm)
   1525     return shm.name, vals.shape, vals.dtype
-> 1527 data_shm = tuple((
   1528     (column, *arr2shm(values))
   1529     for column, values in chain([(Backtest._mp_task_INDEX_COL, self._data.index)],
   1530                                 self._data.items())
   1531 ))
   1532 with patch(self, '_data', None):
...
-> 1521 buf = np.ndarray(vals.shape, dtype=vals.dtype, buffer=shm.buf)
   1522 buf[:] = vals[:]  # Copy into shared memory
   1523 assert vals.ndim == 1, (vals.ndim, vals.shape, vals)

TypeError: buffer is too small for requested array
****
```

As you can see I have removed the optimization ranges and just gave one value per parameter, but it still fails. The original backtest itself, `bt.run()`,  runs fine, and completes in 0.5sec.

I don't know if `bt.optimize()` runs some kind of vectorized calculations, where the data array can somehow end up being too big for it to handle? Can I instead run the optimization sequentially?

### Code sample

```python

```

### Actual behavior

.

_No response_

### Software versions

backtesting==0.6.2

nyxst4ck:
Fixes #1237.

## The bug

`Backtest.optimize(method='grid')` fails with `TypeError: buffer is too small for requested array` whenever the data has a **RangeIndex**, while `bt.run()` on the same data works fine.

Root cause in `SharedMemoryManager.arr2shm` (`backtesting/_util.py`): the shared-memory block is sized with `vals.nbytes`, but RangeIndex is lazy — its `.nbytes` reports the size of the underlying `range` object itself (a constant 132 bytes), not the int64 array it materializes into. The very next line builds `np.ndarray(vals.shape, dtype=vals.dtype.base, buffer=shm.buf)`, which needs `8 × len(index)` bytes. Measured: `pd.RangeIndex(44640).nbytes == 132` vs 357 120 bytes required — so any range-indexed frame longer than 16 rows overflows.

Contrary to the issue title, size is not the real variable — index type is. The "under 10 000 rows works" observations in the thread were almost certainly datetime-indexed runs, and range-indexed data is an explicitly documented input (`Backtest.__init__`: "a monotonic range index (i.e. a sequence of periods)").

## The fix

Size the block from the same `shape`/`dtype.base` pair the buffer view is built with: `vals.size * vals.dtype.base.itemsize`. For every eager index in play (`Index[int64]`, `DatetimeIndex`, tz-aware `DatetimeIndex`, float Series) this equals `.nbytes`, so only the broken RangeIndex path changes behavior.

## Validation

- New regression test `test_optimize_range_index` (placed next to the closely-related `test_optimize_datetime_index_with_timezone`): red on master with the exact `TypeError` from the issue, green with the fix.
- Correctness beyond not-crashing: with a real `multiprocessing.Pool` (not the Windows thread fallback), a 4-point grid on range-indexed GOOG returns the same best equity (81 812.37) as four independently computed serial `bt.run()` calls.
- `python -m backtesting.test`: 81 tests, OK (1 skipped).
- `flake8`: clean. `mypy`: error list byte-identical to master (all pre-existing, none introduced).

### DIFF
--- a/backtesting/_util.py
+++ b/backtesting/_util.py
@@ -309,7 +309,9 @@ class SharedMemoryManager:
     def arr2shm(self, vals):
         """Array to shared memory. Returns (shm_name, shape, dtype) used for restore."""
         assert vals.ndim == 1, (vals.ndim, vals.shape, vals)
-        shm = self.SharedMemory(size=vals.nbytes, create=True)
+        # Not `vals.nbytes`; lazy indexes (e.g. RangeIndex) report their own,
+        # constant size rather than that of the array they materialize into
+        shm = self.SharedMemory(size=vals.size * vals.dtype.base.itemsize, create=True)
         # np.array can't handle pandas' tz-aware datetimes
         # https://github.com/numpy/numpy/issues/18279
         buf = np.ndarray(vals.shape, dtype=vals.dtype.base, buffer=shm.buf)
--- a/backtesting/test/_test.py
+++ b/backtesting/test/_test.py
@@ -1273,6 +1273,16 @@ class TestRegressions(TestCase):
         res = Backtest(data, SmaCross).optimize(fast=range(2, 3), slow=range(4, 5))
         self.assertGreater(res['# Trades'], 0)
 
+    def test_optimize_range_index(self):
+        # Data is passed to the workers through shared memory sized by
+        # `.nbytes`, which a lazy RangeIndex under-reports, previously raising
+        # "TypeError: buffer is too small for requested array". See GH issue #1237.
+        data: pd.DataFrame = GOOG.iloc[:100].reset_index(drop=True)
+        with self.assertWarnsRegex(UserWarning, 'index is not datetime'):
+            bt = Backtest(data, SmaCross)
+        res = bt.optimize(fast=range(2, 3), slow=range(4, 5))
+        self.assertGreater(res['# Trades'], 0)
+
     def test_sl_tp_values_in_trades_df(self):
         class S(_S):
             def next(self):


kernc:
Excellent work! Many thanks!

PR #1388, opened 2026-08-07
ENH: lib: Add deflated_sharpe_ratio() to judge optimize() results

ipezygj:
**What this adds**

`lib.deflated_sharpe_ratio(stats, trial_sharpe_ratios)` — the probability (0–1) that the best run returned by `Backtest.optimize()` has a Sharpe ratio genuinely greater than zero, after correcting for the multiple testing the optimization itself performs.

**Why**

`optimize()` returns the maximum over all tried parameter combinations. The expected best Sharpe of N *skill-less* trials is already well above zero and grows with N, so the winning run's Sharpe is inflated by selection — the more combinations tried, the more the "best" result reflects luck rather than edge. The library currently reports the winner's Sharpe with no way to ask whether it clears that bar. This is the standard correction: Bailey & López de Prado (2014), *The Deflated Sharpe Ratio* ([doi:10.3905/jpm.2014.40.5.094](https://doi.org/10.3905/jpm.2014.40.5.094)).

The trial count *and* the trial Sharpe dispersion are taken from the search itself (the heatmap), not from an assumed default — `optimize()` already produces exactly the data the estimator needs:

```python
stats, heatmap = bt.optimize(fast=range(5, 30, 5), slow=range(10, 70, 10),
                             maximize='Sharpe Ratio', return_heatmap=True)
deflated_sharpe_ratio(stats, heatmap)
```

Concrete example with the test `SmaCross` on `GOOG` (28 combinations): the winner's Sharpe looks significant as a single test (PSR ≈ 0.98) but does **not** clear the hurdle its own 28-trial search sets by chance (DSR ≈ 0.18). That flip is the information the number adds.

**Implementation notes**

- **No new dependencies** — normal CDF/PPF come from stdlib `statistics.NormalDist`; the rest is numpy/pandas already in use.
- The periodic-returns resampling was extracted from `compute_stats()` into `_stats.periodic_returns()` and reused, so the winner's returns are computed by the exact same code path as the reported Sharpe — no duplicated freq logic.
- Warns if the passed heatmap's name isn't `'Sharpe Ratio'` (i.e. optimization maximized something else, so the values wouldn't be trial Sharpes).
- Verified against an independent implementation of the same estimator (agreement to 3+ decimals on the example above and on the single-trial degenerate case, where DSR reduces to the probabilistic Sharpe ratio).
- Unit test included; `flake8` clean; full `python -m backtesting.test` suite passes.


ipezygj:
Pushed a correctness fix to this branch, found by probing my own implementation rather than by review.

A constant return series has no Sharpe ratio, but it did not reach the deflation arithmetic as a `nan`: the standard deviation of a constant series is floating-point residue rather than an exact zero, so a flat 0.1% series divided out to a Sharpe of ~1e16. That is finite, so it passed every guard, and the deflated Sharpe came out **1.0** — certainty of a real edge, from the one input that carries no information about one. The statistic exists to catch results that are too good to be true, so answering with maximum confidence there is the worst possible failure direction.

The guard now compares the dispersion against the resolution of a float at the scale of the data, scaled by the number of terms summed (`n * eps * scale`), rather than against zero. Measured over constant series spanning values 1e-7..1e3 and lengths 3..10000, the residue reaches at most 1.96 eps × scale, so this keeps a margin of at least 3.9x at every length, while a real series with sigma=1e-12 sits more than ten orders of magnitude above the threshold — nothing legitimate is caught, and the test asserts that a quiet series still gets a number.

Two notes on the test, since they are the reason this took a second pass:

- It sweeps values × lengths instead of asserting one series. My first attempt used a single flat series, passed, and still leaked at other lengths — the residue depends on both the value and the length, so a guard calibrated on one point tests only that point.
- It was run against the unfixed function and fails there, so it tests the guard rather than merely accompanying it.

Happy to split this into a separate commit or reword anything if you would rather review it apart from the feature.

kernc:
> The expected best Sharpe of N _skill-less_ trials is already well above zero and grows with N, so the winning run's Sharpe is inflated by selection — the more combinations tried, the more the "best" result reflects luck rather than edge.

That's a reasonable idea.

Can you offer any comparison of Bailey-LdP scoring method against a simpler:
```py
n_trials, sr_mean, sr_std, sr_max = len(heatmap), heatmap.mean(), heatmap.std(), heatmap.max()
penalty = (1 + lamda * np.log1p(n_trials))
score = norm.cdf((sr_max - sr_mean) / sr_std / penalty)
```

ipezygj:
Measured it rather than argued it. Short answer: your penalty is a fine **ranking** device inside one search and nearly as powerful as Bailey–LdP once you match false-positive rates, but it cannot be a **reported probability**, for two structural reasons that the simulation makes visible.

Setup: `N` skill-less parameter combinations, `T` daily returns each, winner picked by max Sharpe afterwards — the exact situation both statistics exist to catch. 2000 independent searches per cell. A score reported as a probability should fire above 0.95 on 5% of these.

### 1. False positives when nothing has an edge

| grid | chance hurdle E[max SR] | DSR | λ=0.1 | λ=0.5 | λ=1 | λ=2 |
|---|---|---|---|---|---|---|
| N=9, T=252 | 1.52 | 0.0% | 8.8% | 0.0% | 0.0% | 0.0% |
| N=25, T=252 | 2.00 | 0.2% | 27.4% | 0.0% | 0.0% | 0.0% |
| N=100, T=252 | 2.53 | 0.1% | 57.9% | 0.0% | 0.0% | 0.0% |
| N=400, T=252 | 2.98 | 0.0% | 84.4% | 0.0% | 0.0% | 0.0% |
| N=100, T=1260 | 1.13 | 0.1% | 58.2% | 0.0% | 0.0% | 0.0% |
| N=100, T=63 | 5.06 | 0.2% | 59.5% | 0.0% | 0.0% | 0.0% |

Two things to read off this.

**λ does not survive a change in N.** At λ=0.1 the false-positive rate *rises* with the number of trials — 8.8% → 84.4% — which is backwards: the more combinations you try, the more confidently it certifies the winner. The max of N draws grows like σ·√(2 ln N) while the penalty grows like 1+λ·log1p(N); over the N range an `optimize()` grid actually spans, the numerator wins. Push λ to 0.5 and it never fires again, in any cell, including the ones below where a real edge is present. There is no λ that is right for both ends.

**The score has no T.** Look at the last three rows: N=100 at 63, 252 and 1260 days. The real chance hurdle moves by a factor of 4.5 (5.06 → 1.13, because a Sharpe estimated from 63 days is far noisier than one from 1260). The heuristic's verdict does not move at all: 59.5%, 57.9%, 58.2%. It cannot move, because its whole input is the heatmap — a three-month backtest and a five-year backtest with the same spread of Sharpes are indistinguishable to it. That is a missing variable, not a tuning problem, and it is where the `√(n_periods − 1)` in Bailey–LdP is doing its work.

### 2. Power, with the calibration question stripped out

The comparison that is fair to your version: fit each score's threshold *per cell* so it fires on exactly 5% of skill-less searches, then plant one combination with a true Sharpe of twice the chance hurdle and see how often each catches it. This gives the heuristic the calibration it cannot compute for itself.

| grid | DSR | λ=0.5 |
|---|---|---|
| N=9, T=252 | 60.1% | 36.6% |
| N=25, T=252 | 81.5% | 72.9% |
| N=100, T=252 | 94.8% | 93.8% |
| N=400, T=252 | 98.9% | 98.8% |
| N=100, T=1260 | 96.0% | 95.5% |
| N=100, T=63 | 93.0% | 91.6% |

So your instinct is right for the large-grid case: at matched FPR they are within a percentage point once N ≥ 100. The gap opens on small grids (60.1% vs 36.6% at N=9), where the winner's own return series carries information the heatmap's three summary numbers throw away. If all you want is to sort runs within a single `optimize()` call, the cheap version very nearly does it.

### 3. What the deflated version buys, and one thing it doesn't

DSR's null distribution is the same in all six cells — median 0.476–0.493, 95th percentile 0.759–0.786 — so 0.95 means the same thing whether the grid had 9 combinations or 400, and whether the backtest was three months or five years. That portability is the whole product; it is what lets the number be printed next to a result and read by someone who did not run the search.

Honest caveat in the other direction: it is *conservative*, not exactly calibrated. Under the null it exceeds 0.95 in 0.0–0.2% of searches, not 5%, and the 99th percentile of the null is 0.86. So the "below ~0.95 suggests overfitting" line in my docstring is a stricter bar than it sounds, and on small grids it will miss real edges (13.8% detection at N=9 at the fixed 0.95 threshold). Happy to reword that guidance, or to expose the threshold, if you'd rather the docstring not imply a 5% test.

Script that produced all of the above, including a parity check of the fast re-implementation against `backtesting.lib.deflated_sharpe_ratio` on the same inputs (max |diff| = 0.00e+00, so the tables are about the shipped function, not a lookalike): happy to attach it here or add it under `tests/` as a slow-marked calibration test if that is useful to you.

